PR #248, opened 2023-07-21
feat: ✨ sd divider

Vahid1919:


packages/components/src/components/divider/divider.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+  parameters: { ...parameters },
+  decorators: [
+    (story: () => typeof html) => html`
+      <style>
+        sd-divider[orientation='vertical'] {
+          height: 120px;
+        }
+        sd-divider[orientation='horizontal'] {
+          width: 120px;
+        }
```
mariohamann, line 23: Please add a story which shows, how the size can influenced by just setting e. g. the width.

Show 2 or 3 sizes and maybe just do it with a template. (Don't do the %TEMPLATE% thing in the constants but just set different values on the y Axis – I believe that should work.)

packages/components/src/components/divider/divider.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @status stable
+ * @since 1.0
+ *
+ * @cssparts base - The component's base wrapper.
+ */
+@customElement('sd-divider')
+export default class SdDivider extends SolidElement {
+  /** Determines the orientation of the divider. */
+  @property({ reflect: true }) orientation: 'horizontal' | 'vertical' = 'horizontal';
+  /** This inverts the divider. */
```
mariohamann, line 19: Should there be a linebreak above that comment?

packages/components/src/components/divider/divider.ts
```diff
@@ -0,0 +1,62 @@
...
+    SolidElement.styles,
+    css`
+      :host {
+        display: block;
+        margin: 0;
+      }
+
+      :host(sd-divider[orientation='vertical']) {
+        display: inline-block;
+      }
```
mariohamann, line 53:
```suggestion
      :host {
        margin: 0;
      }

      :host(sd-divider[orientation='horizontal']) {
        display: block;
      }

      :host(sd-divider[orientation='vertical']) {
        display: inline-block;
      }
```

I think this could improve understandability. :)

packages/components/src/components/divider/divider.stories.ts
```diff
@@ -0,0 +1,155 @@
...
+        y: { type: 'attribute', name: 'inverted' }
+      },
+      args
+    });
+  }
+};
+
+/**
+ * Use a wrapper to set the width and height of the divider.
+ * For horizontal dividers, only set the `width`, and for vertical dividers, adjust the `height`.
```
mariohamann, line 73: The wrapper is only needed here for the story, in the end you're influencing the size of the element itself. You could even set the size with inline styles.
```suggestion
 * You can set the component's size with `width` (horizontal) or `height` (vertical) via CSS or inline-styles.
```

packages/components/src/components/divider/divider.stories.ts
Vahid1919: @mariohamann I added the sizes story but there might be a cleaner way to organise the CSS. Tried organising the story into x and y axis, but I couldn't figure out how to make variations based on orientation and the class names(since that is not a prop and was only needed in the class).

mariohamann:
One last thing from my side: Please set the background to UI blue in for the inverted state in the inverted story. You can have a look in the button inverted story to see, how we handle this.

mariohamann:
- [ ] @Vahid1919 Please check the box, if you made sure that `pnpm verify` still works (including all tests) on your local machine, as we currently can't rely on the GitHub Quality Gate... :(

mariohamann:
@Vahid1919

@EllaELSA approved the PR – you're good to merge (if pnpm verify still works)! :)

mariohamann (approved):

PR #252, opened 2023-07-24
feat: sd-brandshape

christophsaile:


packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+        primary: 'fill-primary/[var(--brandshape-opacity)]',
+        white: 'fill-white/[var(--brandshape-opacity)]'
+      }[this.variant]
+    );
+  }
+
+  connectedCallback(): void {
+    super.connectedCallback();
+    this.resizeObserver = new ResizeObserver(() => {
+      const debounceFunc = debounce(100, () => {
```
mariohamann, line 71: Why don't you directly call the function, as it is only used once?

packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private get bottomSvg(): TemplateResult {
+    return html`<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 958 168">
+      <path
+        class=${this.svgClasses()}
+        d="M121.173 166.106 958 3.216V0H0v66.312c0 6.524.628 13.033 1.876 19.436 10.74 55.114 64.151 91.092 119.297 80.358Z"
+      />
+    </svg>`;
+  }
```
mariohamann, line 56: Was it a problem to use the topSvg twice by flipping?

packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  private svgClasses(): string {
+    return cx(
+      {
+        'neutral-100': 'fill-neutral-100/[var(--brandshape-opacity)]',
+        primary: 'fill-primary/[var(--brandshape-opacity)]',
+        white: 'fill-white/[var(--brandshape-opacity)]'
```
mariohamann, line 63: Can you use all transparent variants from a brand perspective?

Still, I believe we need either more variants e. . `primary-opaque`, `neutral-100-opaque` or have a boolean `opaque' if this really is for every bg-color.

Background: Brand and Design system define the allowed opacity (AFAIK), not the user of the component.

packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @cssproperty --brandshape-opacity - The opacity of the brandshape.
+ */
+@customElement('sd-brandshape')
+export default class SdBrandshape extends SolidElement {
+  /** The brandshape's theme variant. */
+  @property({ type: String }) variant: 'neutral-100' | 'primary' | 'white' = 'primary';
+
+  /** Defines the form of the brandshape  */
+  @property({ type: String }) form: 'full' | 'top' | 'topCenter' | 'centerBottom' = 'full';
```
mariohamann, line 32: Question: Wouldn't it be possible to do this with either CSS parts or CSS properties?

```
sd-brandshape{
     --top: none;
     --middle: none;
}
```

```
sd-brandshape::part(top){
display: none;
}
```

This could make it possible to maybe even remove a bunch of JS, wouldn't it?

Furthermore:
Let's call it top, middle, bottom, as center is usually used for horizontal alignment and middle for vertical alignment.

If you have to go the JS route, at least I would convert this thing to an array, that people can write `top,middle` or `top` and `top,middle,bottom` is the default. See https://lit.dev/docs/components/properties/#conversion-type

Your solution might work for CMS, but we don't know what happens in future and we don't want to add a bunch of properties.

mariohamann (approved):

packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+      top: [this.renderTopBrandshape(true)]
+    };
+  }
+
+  private renderTopBrandshape(includeSlot?: boolean): TemplateResult {
+    return includeSlot
+      ? html` <div class="relative">
+          ${this.topSvg}
+          <div part="content-top" class="absolute bottom-0 right-0 flex items-end w-2/5 h-2/3 px-6 py-4">
+            <slot></slot>
```
mariohamann, line 102: How does this work if there are two main slots?

Please make this a named slot.

And please add stories for both.

packages/components/src/components/brandshape/brandshape.ts
```diff
@@ -0,0 +1,155 @@
...
+      centerBottom: [this.renderCenterBrandshape(), this.renderBottomBrandshape()],
+      top: [this.renderTopBrandshape(true)]
+    };
+  }
+
+  private renderTopBrandshape(includeSlot?: boolean): TemplateResult {
+    return includeSlot
+      ? html` <div class="relative">
+          ${this.topSvg}
+          <div part="content-top" class="absolute bottom-0 right-0 flex items-end w-2/5 h-2/3 px-6 py-4">
```
mariohamann, line 101: Couldn't we make it that the text follows more or less the form? (A simple triangle would be enough.)

https://developer.mozilla.org/en-US/docs/Web/CSS/shape-outside

packages/components/src/components/brandshape/brandshape.ts
christophsaile: The viewbox of the top svg is different from the one of the bottomSvg. I Thought the code is more understandable this way. But yeah inceases file size a bit.

packages/components/src/components/brandshape/brandshape.ts
christophsaile: copied form the docs :D, fixed

packages/components/src/components/brandshape/brandshape.ts
christophsaile: There are never two main slots, you can only put in content into the top svg if variant === top. Therefore I think a named slot would be unnecessary. You should only be able to add one content

packages/components/src/components/brandshape/brandshape.ts
christophsaile: In the cms-context we need to be able to set the following opacities : 80%, 90%, 100% on white and primary. We could also make this more strict so only this 3 options are available

packages/components/src/components/brandshape/brandshape.ts
christophsaile: I will give that a try

packages/components/src/components/brandshape/brandshape.ts
christophsaile: mhhh I don't think it would reduce the amount of js that much since most of the js is just rendering template which you would need with css aswell. Also this means that even thought you only selected top variant, the whole brandshape would rendered in the dom.

Top, middle, bottom sounds good.

I like the array solution, this also matches my current rendering approach.

packages/components/src/components/brandshape/brandshape.ts
mariohamann: Okay, this should be definitely more strict, that doesn't sound very systematic.

Maybeeeee... as a solution for the moment you could just set the color and override the opacity with the part selector and pointing to the SVG?

I believe we will introduce opacity later in some kind of property, but maybe for the moment to get you running in your project this would be easier?

mariohamann:
@van-nguyen-ht @EllaELSA Please review.

We focused only on the full colored variants. Transparent background colors and outline variants will follow.

EllaELSA:
the rounding of brand shape behind the claim is very big. This results in a very high shape element (feels very dominant and covers too much of the image). Would be better to not use the same rounding as in the full shape Version behind the text but have a second rounding element with a smaller rounding for the claim version. 
spoke about this with @christophsaile and will add this as an improvement of this element

mariohamann:
@EllaELSA @van-nguyen-ht Could one of you please check out the Chromatic tests?

van-nguyen-ht:
looking good 
here's the parallel iteration in Figma 
https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/branch/s7zl5cMsCXzG0gp670AFmu/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=8713-2967&mode=design&t=q0hqtnJmsURsumBX-11

PR #268, opened 2023-07-26
ci: purge all folders after deployment (hotfix)

mariohamann:
Currently the purging of the CDN is not working:

CleanShot 2023-07-26 at 16 40 58@2x

I would propose a hotfix for now that purges just everything and create a followup ticket to make it clean.

### DIFF
--- a/.azuredevops/pipelines/templates/purge-cdn.template.yml
+++ b/.azuredevops/pipelines/templates/purge-cdn.template.yml
@@ -7,43 +7,53 @@ steps:
       scriptLocation: 'inlineScript'
       inlineScript: |
         if [[ $deployment != "none" ]]; then
-          # Purge version folder
+          # Purge all folders
           set -x
           az cdn endpoint purge \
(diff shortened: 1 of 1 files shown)


karlbaumhauer:
seems like there is something more not working: 
image


is this really critical? the CDN purges itself every night (AFAIK)
otherwise I would suggest to dig deeper into why the version numbers are not set... (which seems to be the reason)

karlbaumhauer (approved):

mariohamann:
Chris was dependent on a new feature branch version for CMS + we had problems sometimes on Storybook with old cached data. I suspect the latter could be the same problem

mariohamann:
If you have time of course directly fixing it would be better @karlbaumhauer ? :)

karlbaumhauer:
@mariohamann I will have a deeper look into it... this situation makes me nervous about the releases...;-)

karlbaumhauer:
@mariohamann @christophsaile I think I found the issue and created a new PR => #290

PR #286, opened 2023-07-27
feat: ✨ sd teaser

azraefendic:


packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,163 @@
...
+
+          <div
+            part="headline"
+            class=${cx(
+              'text-lg font-bold mb-1',
+              this.variant === 'primary' ? 'text-white' : 'text-black',
+              !slots['headline'] && 'hidden'
+            )}
+          >
+            <slot name="headline"></slot>
```
mariohamann, line 140:
```suggestion
            <slot name="headline">Always insert one semantically correct heading element here &lt;h&gt </slot>
```

As we said: We should provide a default value in the component, not in the story. :) Then everyone using the component without looking to Storybook sees it.

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+        y: {
+          type: 'template',
+          name: 'media and content distribution',
+          values: [
+            {
+              title: 'media: 66%, content: 33%',
+              value: '<div style="--distribution-media: 66%; --distribution-content: 33%;">%TEMPLATE%</div>'
+            },
+            {
+              title: 'media: 200px, content: 450px',
```
mariohamann, line 119: would content 100% work? I think it's more realistic to have only one fixed value and let the other one resize automatically.

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,163 @@
...
+@customElement('sd-teaser')
+export default class SdTeaser extends SolidElement {
+  @property({ reflect: true })
+  variant: 'white' | 'white border-neutral-300' | 'neutral-100' | 'primary' | 'primary-100' = 'white';
+
+  /** The teaser's orientation */
+  @property({ reflect: true })
+  orientation: 'responsive' | 'vertical' | 'horizontal' = 'responsive';
+
+  /** The teaser's padding */
```
mariohamann, line 38:
```suggestion
  /** The teaser's inner padding. This is always set in `white border-neutral-300`. */
```

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,163 @@
...
+      headline: this.hasSlotController.test('headline')
+    };
+
+    return html`
+      <div
+        class=${cx(
+          'flex',
+          {
+            white: 'bg-white',
+            'white border-neutral-300': 'bg-white border border-neutral-300',
```
mariohamann, line 91: That's a good idea to split this with a space! So the first one is always the background and then comes the border.

When splitting it with a space, we could even name name it `white neutral-300` like design proposed – what do you prefer?

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+    },
+    {
+      type: 'slot',
+      name: 'media',
+      value: `<slot-comp slot='media' style="--slot-height: 48px; --slot-content: 'Media slot'"></slot-comp>`
+    },
+    {
+      type: 'slot',
+      name: 'headline',
+      value: '<h2 slot="headline">Remember: Always Use &lt;h&gt; Tags here!</h2>'
```
mariohamann, line 26: Remove that in favor of the default slot as proposed belowl

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+    },
+    {
+      type: 'slot',
+      name: 'meta',
+      value: `<slot-comp slot='meta' style="--slot-height: 46px; --slot-content: 'Meta slot'"></slot-comp>`
+    }
+  ]),
+  argTypes,
+  parameters: { ...parameters }
```
mariohamann, line 35:
```suggestion
  parameters
```

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+  render: (args: any) => {
+    return generateTemplate({ args });
+  }
+};
+
+/**
+ * Teaser in all possible variants.
+ */
+
+export const Variants = {
```
mariohamann, line 52: Maybe you could extend it to `VariantAndInset` to see, that the thing with the border has always inset.

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+export const Headline = {
+  parameters: { controls: { exclude: 'headline' } },
+  render: (args: any) => {
+    return generateTemplate({
+      axis: {
+        y: {
+          type: 'slot',
+          name: 'headline',
+          values: [
+            {
```
mariohamann, line 206: Please define the more programmatic, that something like 
```js
['h1', 'h2', 'h3'].map((tag) => { return {value: `<${tag} slot="headline">${tag}</${tag}`, title: tag}})
```
or similar (untested)

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,352 @@
...
+          name: 'orientation',
+          value: 'vertical'
+        }
+      ]
+    });
+  }
+};
+
+/**
+ * Teaser in all possible combinations of `inset` and 'orientation'. Note that in case the variant is `white border-neutral-300`, the inset is always `true`.
```
mariohamann, line 77: I would write this then at `VariantAndInset` as this belongs to that.

Furthermore:

CleanShot 2023-07-27 at 17 16 01@2x

The elements should all have the same width and height, so that we can compare how they behave differently in regard to their properties. :)

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        <div
+          style=${this.orientation === 'horizontal' || this.currentOrientation === 'horizontal'
+            ? `width: var(--distribution-content, 100%); ${
+                this.inset ? 'width: var(--distribution-content, calc(100% - 2rem));' : ''
+              }`
+            : ''}
+          class=${cx(
+            'flex flex-col text-left',
+            (this.orientation === 'horizontal' || this.currentOrientation === 'horizontal') && `flex flex-col`,
```
mariohamann, line 123: Here's too much `flex` and `flex-col`, they should appear only once.

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,163 @@
...
+          {
+            white: 'bg-white',
+            'white border-neutral-300': 'bg-white border border-neutral-300',
+            'neutral-100': 'bg-neutral-100',
+            primary: 'bg-primary text-white',
+            'primary-100': 'bg-primary-100'
+          }[this.variant],
+          (this.orientation === 'vertical' || this.currentOrientation === 'vertical') && 'flex-col',
+          (this.orientation === 'horizontal' || this.currentOrientation === 'horizontal') && 'flex-row gap-8',
+          (this.orientation === 'horizontal' || this.currentOrientation === 'horizontal') && this.inset && 'py-8 px-10'
```
mariohamann, line 98: You could always use `currentOrientation` if you set it once when initiating the component. It shoudl work in the constructor phase.

packages/components/src/components/teaser/teaser.ts
azraefendic: Hmm, was it proposed in design? This is the name I got from figma, I feel like it makes sense because it makes it clear that this variant has a border unlike the others, but I guess both work!

mariohamann: If it makes it more clear, let's leave it like it is. :)

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,164 @@
...
+        >
+          <div part="meta" class=${cx('gap-2 mb-4', !slots['teaser-has-meta'] && 'hidden')}>
+            <slot name="meta"></slot>
+          </div>
+
+          <div
+            part="headline"
+            class=${cx('text-lg font-bold mb-1', this.variant === 'primary' ? 'text-white' : 'text-black')}
+          >
+            <slot name="headline">Always insert one semantically correct heading element here &lt;h&gt;</slot>
```
mariohamann, line 141:
```suggestion
            <slot name="headline">Always insert one semantically correct heading element here (e. g. &lt;h2&gt;)</slot>
```

Sorry, I missed that. :D

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,164 @@
...
+          )}
+          id="content"
+        >
+          <div part="meta" class=${cx('gap-2 mb-4', !slots['teaser-has-meta'] && 'hidden')}>
+            <slot name="meta"></slot>
+          </div>
+
+          <div
+            part="headline"
+            class=${cx('text-lg font-bold mb-1', this.variant === 'primary' ? 'text-white' : 'text-black')}
```
mariohamann, line 139: CleanShot 2023-07-28 at 09 37 35@2x

We have a problem here. 😭

The headlines change their size – and that shouldn't be. They hould have the same size for every headline style.

We have two possibilities:
- set important on the classes https://tailwindcss.com/docs/configuration#important-modifier – but I'm not sure if this will cascade to the slot
- work with ::slotted e. g. https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted#try_it 
```
::slotted([slot=headline]) {
    /* text-size boldness */
}
```

packages/components/src/components/teaser/teaser.stories.ts
mariohamann: No, this is totally fine like this.

packages/components/src/components/teaser/teaser.stories.ts
```diff
@@ -0,0 +1,338 @@
...
+  parameters: { controls: { exclude: ['--distribution-media', '--distribution-content'] } },
+  render: (args: any) => {
+    return generateTemplate({
+      axis: {
+        y: {
+          type: 'template',
+          name: 'media and content distribution',
+          values: [
+            {
+              title: 'media: 33%, content: 66%',
```
mariohamann, line 117:
```suggestion
              title: '--distribution-media: 33%, --distribution-content: 66%',
```

mariohamann: I think it's better to show the CSS variable in the title.

packages/components/src/components/teaser/teaser.ts
```diff
@@ -159,6 +159,12 @@ export default class SdTeaser extends SolidElement {
       ::slotted(*) {
         margin: 0;
       }
+
+      ::slotted([slot='headline']) {
+        font-size: 20px !important;
+        font-weight: 700px;
```
mariohamann, line 165: Please set `bold`. (And it would be 700 not 700 px.)

mariohamann: Or do we have 700 everywhere? Hmhmhmm.

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,170 @@
...
+      :host {
+        display: block;
+      }
+
+      ::slotted(*) {
+        margin: 0;
+      }
+
+      ::slotted([slot='headline']) {
+        font-size: 20px !important;
```
mariohamann, line 164:
```suggestion
        font-size: var(--sd-font-size-lg, 1.25rem) !important;
```

mariohamann: This should still be overridable via CSS variables.

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,170 @@
...
+  updated(changedProperties: PropertyValues) {
+    super.updated(changedProperties);
+    if (changedProperties.has('orientation')) {
+      this.updateOrientation();
+    }
+  }
+
+  updateOrientation() {
+    if (this.orientation === 'responsive') {
+      this.currentOrientation = this.offsetWidth > 448 ? 'horizontal' : 'vertical';
```
mariohamann, line 74: I'm just realizing: The component is missing the `breakpoint` property. The 448px should be overridable from outside.

Thinking about that... do you think we would need the `responsive` property as someone could just set the `breakpoint` to `0` for always vertical and to e. g. `9999` for always horizontal? Should we do a session on that together? :)

packages/components/src/components/teaser/teaser.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      const media = el.shadowRoot!.querySelector<HTMLElement>('[part="media"]');
+      const content = el.shadowRoot!.querySelector<HTMLElement>('[part="content"]');
+
+      if (media && content) {
+        const mediaWidth = media.offsetWidth;
+        const contentWidth = content.offsetWidth;
+
+        // Here you have the widths, you can do something with them
+        console.log(`Media width: ${mediaWidth}, Content width: ${contentWidth}`);
```
mariohamann, line 120: ChatGPT? 😆

packages/components/src/components/teaser/teaser.test.ts
```diff
@@ -0,0 +1,126 @@
...
+describe('<sd-teaser>', () => {
+  it('renders default values correctly', async () => {
+    const el = await fixture<SdTeaser>(html`<sd-teaser></sd-teaser>`);
+
+    expect(el.variant).to.equal('white');
+    expect(el.breakpoint).to.equal(448);
+    expect(el.inset).to.equal(false);
+  });
+
+  it('renders assigned values correctly', async () => {
```
mariohamann, line 13: We need tests for the breakpoints.

Something like:

default breakpoint
1. set width of component < breakpoint
2. check if state is vertical
3. set width of component > breakpoint
4. check if state is horizontal

manual breakpoint
1. set width of component < breakpoint
2. check if state is vertical
3. set width of component > breakpoint
4. check if state is horizontal

mariohamann:
## Migration guide

We have to add the migration guide for the component below... maybe in another PR? What do you think @azraefendic ?

```js
  @Prop() headline: string;

  @Prop() text: string;

  @Prop() date: string;

  @Prop() category: string;

  @Prop() author: string;

  @Prop() image: string;

  /**
   * Provides the original asset ratio to be considered in the responsive image URL.
   * See ui-image component for more information.
   */
  @Prop() imageRatio?: string;

  @Prop() icon: Icon;

  @Prop({ attribute: 'image-alt-text' }) imageAltText: string;

  @Prop({ attribute: 'image-mouseover' }) imageMouseover: string;

  @Prop() copyright: string;

  @Prop() alignButton = false;

  @Prop({ attribute: 'media-selection' }) mediaSelection: 'none' | 'image' | 'video' | 'icon';

  @Prop({ attribute: 'video-src' }) videoSrc: string;

  @Prop({ attribute: 'video-loop-setting' }) videoLoopSetting?: string;

  @Prop() replace?: string;

  @Prop() sizes?: string;

  @Method()
  async hasLinks() {
    return this.links?.length > 0;
  }

  @Prop({ mutable: true }) links!: Link[];

  @Prop({ attribute: 'links' }) utilityLinks: string;
```

azraefendic:
Yes, I would add the migration guide in another PR then @mariohamann.

Vahid1919 (approved):

packages/components/src/components/teaser/teaser.ts
```diff
@@ -0,0 +1,167 @@
...
+      :host {
+        display: block;
+      }
+
+      ::slotted(*) {
+        margin: 0;
+      }
+
+      ::slotted([slot='headline']) {
+        font-size: var(--sd-font-size-lg, 1.25rem) !important;
```
yoezlem, line 161: Why do we need the !important here? Can't we get rid of that somehow? 🥲

azraefendic: I know 🥲 Unfortunately without it it does not override the slotted headline styles, and we need it to look the same regardless of the <h> tag used in the slot.

mariohamann: I think this is okay, as this should really be the size which design defined in the card. It shouldn't be changable.

yoezlem: @mariohamann is this the expected behaviour or is something wrong with the order of overriding in our main file for CSS?

yoezlem: Okay, your comment was not loaded. Good, than its okay for me.

yoezlem:
Code Review LGTM!

yoezlem (approved):

packages/components/src/components/teaser/teaser.ts
mariohamann: The problem is: We don't know how or even if people override/reset their CSS for headlines. And they really shouldn't have to care about when using the component.

PR #328, opened 2023-08-10
docs: 📚 Adding tokens and a11y to CONTRIBUTING.md

MarcMatthiae:
## Description:
I slightly changed the "How we work" section. I actually wanted it to be all "We..."-sentences, but couldn´t find a way to formulate "Development closely works with Design." that way.
I added the usage of tokens and the optimization for accessibility to our dev guidelines, because I think they are important topics.

## Definition of Reviewable:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,9 @@
 - Packages have to be run individually (eg: `cd packages/components` → `pnpm dev` to start development server)
 - Run `pnpm verify` at the root directory periodically, particularly, before pushing changes when a pull request is already opened.
 - Every branch should be associated with a PR.
+- Nearly all styles (colors, fonts, sizes etc.) are defined by our design team and provided for usage in our code via tokens (/packages/tokens/src/token.json).
+  Components should use this tokens instead of individual styles if possible.
+- Components should be optimized for accessibility. Check https://www.w3.org/WAI/standards-guidelines/wcag/ for more information on accessibility.
 
 ## Pull Requests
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -32,6 +33,9 @@
 - Packages have to be run individually (eg: `cd packages/components` → `pnpm dev` to start development server)
 - Run `pnpm verify` at the root directory periodically, particularly, before pushing changes when a pull request is already opened.
 - Every branch should be associated with a PR.
+- Nearly all styles (colors, fonts, sizes etc.) are defined by our design team and provided for usage in our code via tokens (`packages/tokens/src/token.json`).
+  Components should use this tokens instead of individual styles as much as possible.
```
karlbaumhauer, line 37:
```suggestion
  Components should use this tokens instead of individual styles as much as possible. Only where the tokens do not provide a styling, component specific styles should be added inside the [component-name].ts file
```

karlbaumhauer (changes requested):

karlbaumhauer (approved):

CONTRIBUTING.md
```diff
@@ -32,6 +33,9 @@
 - Packages have to be run individually (eg: `cd packages/components` → `pnpm dev` to start development server)
 - Run `pnpm verify` at the root directory periodically, particularly, before pushing changes when a pull request is already opened.
 - Every branch should be associated with a PR.
+- Nearly all styles (colors, fonts, sizes etc.) are defined by our design team and provided for usage in our code via tokens (`packages/tokens/src/token.json`).
+  Components should use this tokens instead of individual styles as much as possible. Only where the tokens do not provide a styling, component specific styles should be added inside the [component-name].ts file
```
azraefendic, line 37: 'this tokens' doesn't sound right, maybe replace it with 'these tokens' or just 'tokens'?

MarcMatthiae: @azraefendic Done :)

azraefendic (approved):

PR #332, opened 2023-08-14
docs: improve opengraph and browser meta appearance

karlbaumhauer:
## Description:
Provide meta information for Storybook pages.

As there is no clear solution to provide basic meta information (especially for the dynamic page title), the following changes help to have some basic mata tags:
- add meta tags to manager-head.html for search enginges, Facebook and Twitter
- rename paths in story fiels to prefix Docs, Components and Uitilities with "SDS" => improve the dynamic title

- closes https://github.com/solid-design-system/internal/issues/28

## Definition of Reviewable:

## Definition of Mergable:

- [x] replace `feature branch` path with `latest` path for favicons and images


karlbaumhauer:
unfortunately the title (which is the most important thing) does not work but at least it shows the favicon and the dynamic title is more descriptive => https://checkserp.com/serp-simulator/ and add https://solid-design-system.fe.union-investment.de/docs_add-meta-information/storybook/index.html

karlbaumhauer:
> unfortunately the title (which is the most important thing) does not work but at least it shows the favicon and the dynamic title is more descriptive => https://checkserp.com/serp-simulator/ or https://metatags.io/ and add https://solid-design-system.fe.union-investment.de/docs_add-meta-information/storybook/index.html

Title works now as well:
image

karlbaumhauer:
@coraliefeil / @van-nguyen-ht if you can provide a signet (with whitespace above and below) in the recommended size, I could even use a better quality... ;-) 

The internet says:
image

packages/components/.storybook/manager-head.html
```diff
@@ -1 +1,123 @@
...
+/>
+<link
+  rel="apple-touch-icon"
+  sizes="72x72"
+  href="https://solid-design-system.fe.union-investment.de/docs_add-meta-information/storybook/favicon/apple-icon-72x72.png"
+/>
+<link
+  rel="apple-touch-icon"
+  sizes="76x76"
+  href="https://solid-design-system.fe.union-investment.de/docs_add-meta-information/storybook/favicon/apple-icon-76x76.png"
```
mariohamann, line 23: We should avoid fixed URLs in anyway, especially to external CDN resources etc.

We're already using dynamic URLs here – doesn't that work here and if not, how do they resolve?

https://github.com/solid-design-system/solid/blob/e1c4e606a5bf4c65658445877ba4a5d19a63e624/packages/components/.storybook/preview-head.html#L16

https://github.com/solid-design-system/solid/blob/e1c4e606a5bf4c65658445877ba4a5d19a63e624/packages/components/.storybook/preview-head.html#L16

packages/components/src/_components/animated-image/animated-image.stories.ts
```diff
@@ -6,7 +6,7 @@ const { argTypes, args, parameters } = storybookDefaults('sd-animated-image');
 const { generateTemplate } = storybookTemplate('sd-animated-image');
 
 export default {
-  title: 'Components/sd-animated-image',
+  title: 'SDS Components/sd-animated-image',
```
mariohamann, line 9: I believe the problem is that the Site Title is generated via JS here, so it works in the browser – but is not picked up by Open Graph Stuff. I would therefore strongly suggest to remove the SDS here everywhere, as it just bloats the docs and we furthermore talk about "Solid Components" and not "Solid Design System Components".

CleanShot 2023-08-15 at 07 07 02@2x

packages/components/scripts/storybook/node-modify-page-title.cjs
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Change the generated index.html document title to "Solid Design System by Union Investment | Version: x.x.x".
```
mariohamann, line 5: I get the intention of the change and like the output – but this might lead to problems:

Our x.x.x URL resolves to different versions, but OG always picks up the version of the moment being sent, so when we link something on the latest branch, it might always stay at "1.4.2" – and people will say "ah, I thought this is only this version".

That might lead to confusion – and there already is often enough confusion around those Storybook URLs, I'm sure, that it will just increase...

For fixed Version URLs, this is great, for sure.

----

Therefore: I would suggest to NOT rewrite this for Open Graph on Build Time but only on Runtime via Vite + JS and even saying "Latest Storybook" in the Browser instead of the version on this particular URL. So people see it in the browser always correct (1.2.4 Docs, 1.2.x docs, latest docs etc.)

packages/components/.storybook/manager-head.html
karlbaumhauer: I tried with different ways to avoid fixed URLs but for some reason it only works for the favicon files. The meta tags will resolve the relative URL into `https://solid-design-system.fe.union-investment.de/sds-signet_meta.png`.

As I did not want to mix the URL type in this file I just used to latest branch CDN... If you have any idea how to solve this, I am happy to use relative URLs.

packages/components/src/_components/animated-image/animated-image.stories.ts
karlbaumhauer: As the page title is dynamic in Storybook and there have been lots of discussions if and how to solve this (and most of them seemed to be very hacky), I just tried to improve it a bit for eg. the title shown in tabs in the browser.

Currently it shows the "chapter" and then the source tree of the file the user is in: 
image

By changing it to SDS, it at least shows that the page is part of the SDS:
image

image

I dont see where it bloats the docs here as we just added a prefix to the name but I see the point to change it to something like "Solid Docs" and "Solid Components".

packages/components/scripts/storybook/node-modify-page-title.cjs
karlbaumhauer: I see the point with the wildcard folders. TBH I did not think about them at this point. I will have a look on it and see if I find a solution to rewrite the title according to the version the link contains (fixed versions, latest and wildcard folders).

In general, this title is only used for sharing purpose as Storybook rewrites the title on runtime anyways.

coraliefeil:
@karlbaumhauer og images -> https://github.com/solid-design-system/figma/issues/55

packages/components/.storybook/manager-head.html
mariohamann: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base#open_graph

This seems to be an Open Graph-Problem. So before merging this, I would suggest to point this to our x.x.x branch but leave the favicon stuff etc as it was as only Open Graph needs absolute paths.

packages/components/scripts/storybook/node-modify-page-title.cjs
mariohamann: If this is only for sharing purposes we would have to rewrite it before/for every CDN folder individually.

I definitely wouldn't recommend the current solution but would rather create an issue for that, so we can have a concept, that helps without possible confusion.

karlbaumhauer: not for every folder but for the wildcard folder and as stated before, I did not have them in mind and was not able to find a way to get this done.

it would work for latest, fixed version and feature branches but not for the wildcard folders. Unfortunately this is probably the more important ones next to latest...

packages/components/.storybook/manager-head.html
karlbaumhauer: yes. The plan was to change it back to x.x.x as soon as I merge it. The feature branch is just for testing reasons.

I thought of only having it for the og:tags but then decided to rather have one way of using links to prevent confusion. An alternative could be a comment but I did not want to write a comment into the storybook head to explain why we mixed the URL types. 

Any idea?

mariohamann: Well, for Favicons the version before was correct, for OG apparently the other one. So I would rather write a comment instead of altering to a less correct version.

packages/components/package.json
```diff
@@ -51,7 +51,9 @@
     "build.package": "pnpm build.package/create-index && vite build --config vite.config.package.js && pnpm build.package/remove-index",
     "build.package/create-index": "node scripts/node-generate-package-index.cjs && echo '✅ Index created'",
     "build.package/remove-index": "node -e \"require('fs').unlinkSync('src/solid-components.package.ts')\" && echo '✅ Index removed'",
-    "build.storybook": "storybook build --output-dir dist/storybook && cp dist/custom-elements.json dist/storybook/custom-elements.json",
+    "build.storybook": "storybook build --output-dir dist/storybook && cp dist/custom-elements.json dist/storybook/custom-elements.json && pnpm build.storybook-page-title && pnpm build.storybook-favicon",
+    "build.storybook-page-title": "node scripts/storybook/node-modify-page-title.cjs",
+    "build.storybook-favicon": "node scripts/storybook/node-modify-favicon.cjs",
```
mariohamann, line 56: I would strongly suggest to just write one single command, like `modify-storybook` or similar. Who knows what will come.

As it is a subtask I furthermore suggest to write the command as `build.storybook/modify-storybook`, as this is not a build task that works on is own but always together with `build.storybook`.

packages/components/src/_components/animated-image/animated-image.stories.ts
mariohamann: As you want only to change what is shown in the browser I found a possible solution:



Replace every occurance of the following in all files `chunks*.js` in the folder sb-manager.

```
\u22C5 Storybook
```

We could alter the word `Storybook` to `Solid Design System (Storybook)` instead of appending SDS everywhere.

packages/components/.storybook/assets/favicon/android-icon-192x192.png
mariohamann, line 1: Was it on purpose that you removed the slight rounding and padding of the favicon which was there before? This feels quite crowded in the URL now.

karlbaumhauer: I did not change anything on purpose... I just used another website to generate the favicons as the first one did not create all device type formats.

But I cant see any problems with the new one? 
image
what exactly do you mean? can you send a screenshot?

mariohamann: Ah it's more in Dark Mode that the signet misses some "air to breathe". But nothing blocking from my side . :)

packages/components/src/components/include/include.stories.ts
```diff
@@ -48,7 +48,7 @@ const { generateTemplate } = storybookTemplate('sd-include');
  */
 
 export default {
-  title: 'Utilities/sd-include',
+  title: 'SDS Utilities/sd-include',
```
mariohamann, line 51:
```suggestion
  title: 'Utilities/sd-include',
```

packages/components/src/docs/General/5-Customization.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta } from '@storybook/blocks';
 
-<Meta title="Docs/General/Customization" />
+<Meta title="SDS Docs/General/Customization" />
```
mariohamann, line 3:
```suggestion
<Meta title="Docs/General/Customization" />
```

packages/components/.storybook/assets/favicon/android-icon-192x192.png
karlbaumhauer: follow up task if the design wants it to be changed... ;-)

karlbaumhauer:
@mariohamann I was able to solve most of the issues you mentioned above. 
The "open" points are:
1.  Quality of the favicon and og:images => https://github.com/solid-design-system/figma/issues/108
2. Version in the title => https://github.com/solid-design-system/solid/issues/351

Pls merge the PR if you agree with the solutions.

mariohamann (approved):

PR #353, opened 2023-08-28
feat: ✨ sd-dropdown

mariohamann:


packages/components/src/components/dropdown/dropdown.ts
```diff
@@ -57,6 +53,16 @@ export default class SdDropdown extends SolidElement {
...
+  @property({ type: Boolean, reflect: true }) rounded = false;
+
+  /**
+   * Indicates whether or not the dropdown should automatically resize its content's width/height regarding the available space on screen.
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'no-auto-size' }) noAutoSize = false;
+
   /**
    * The preferred placement of the dropdown panel. Note that the actual placement may vary as needed to keep the panel
    * inside of the viewport.
```
MarcMatthiae, line 68: Regarding the placement attribute: Should it be possible to leave this empty? Because currently I can select an empty string in storybook. Not sure, just asking.

packages/components/src/components/popup/popup.stories.ts
```diff
@@ -5,16 +5,19 @@ import { withActions } from '@storybook/addon-actions/decorator';
 const { argTypes, args, parameters } = storybookDefaults('sd-popup');
 const { generateTemplate } = storybookTemplate('sd-popup');
 
+/**
+ * Popup is a utility that lets you declaratively anchor "popup" containers to another element.
+ */
+
 export default {
```
MarcMatthiae, line 12: I get an error in the console everytime I open this story, because there is no anchor element.
May we can provide sth. per default?

image

mariohamann: Sounds good! I'll do.

packages/components/src/components/dropdown/dropdown.ts
mariohamann: If you do, the default will be used. So it's fine.

packages/components/src/components/dropdown/dropdown.stories.ts
```diff
@@ -0,0 +1,303 @@
...
+      args
+    })
+};
+
+/**
+ * The preferred placement of the dropdown can be set with the `placement` attribute.
+ * Note that the actual position may vary to ensure the panel remains in the viewport.
+ */
+
+export const Placement = {
```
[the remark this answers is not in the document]
mariohamann: Unfortunately not. :D This is like the dropdown behaves, that it looks for a way to show up. Otherwise we can't do the screenshot tests. if you have an idea, I would be happy to resolve that? :)

packages/components/src/components/dropdown/dropdown.stories.ts
```diff
@@ -0,0 +1,303 @@
...
+      args
+    });
+  }
+};
+
+/**
+ * sd-dropdowns and their content are fully accessibile via keyboard.
+ */
+
+export const Mouseless = {
```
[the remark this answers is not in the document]
karlbaumhauer: second that.

in addition to this:
1. the mouse is still visible in the mouseless story and therewith it still has a hover state
2. if I click on the story "Mouseless" I still have the focus on the side bar => not possible to navigate with the keyboard to the trigger.
3. Even after tabbing a while, I end up on the trigger but it has no focus style.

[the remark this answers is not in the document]
mariohamann: Oh thanks for pointing that out again... I was aware of it – but now finally realized, that it was really a bug in the dropdown component that I could fix. Will push a fix soon.

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
karlbaumhauer: strange... the indicator are completely gone now...

packages/components/src/components/dropdown/dropdown.stories.ts
```diff
@@ -0,0 +1,303 @@
...
+      })}
+    </div>`;
+  }
+};
+
+/**
+ * For round triggers, set the `rounded` attribute. this automatically adds a minimal distance of 1px between the trigger and the panel.
+ */
+
+export const Rounded = {
```
karlbaumhauer, line 89: is it supposed to be that the trigger is rounded as well?
it feels confusing as this is nothing we provide but control it with this prop.

packages/components/src/components/dropdown/dropdown.stories.ts
```diff
@@ -0,0 +1,303 @@
...
+      },
+      args
+    })
+};
+
+/**
+ * The overlay automatically adjusts its size to fit the screen. The content gets scrollable then. To disable this behavior, set the `no-auto-size` attribute.
+ */
+
+export const NoAutoSize = {
```
karlbaumhauer, line 112: in the docs page it looks very broken:
Bildschirmfoto 2023-08-29 um 17 13 14

packages/components/src/docs/Migration/ui-dropdown.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+### 🔀 Updated Attribute: `placement`
+
+The old `dropdownPosition` has been replaced by the `placement` attribute. Here's how you can map the values:
+
+- `bottom-left` in `ui-dropdown` translates to `bottom-start` in `sd-dropdown`.
+- `bottom-right` in `ui-dropdown` translates to `bottom-end` in `sd-dropdown`.
+- on top you have the possibility to set `top`, `left`, `right` including `*-start` and `*-end` variants, too.
+
+## ✨ New Attributes in `sd-dropdown`
```
karlbaumhauer, line 35:
```suggestion
## ✨ New Attributes
```

packages/components/src/docs/Migration/ui-dropdown.mdx
```diff
@@ -0,0 +1,62 @@
...
+- `disabled`: Disables the dropdown.
+- `stayOpenOnSelect`: Keeps the dropdown open after an item is selected.
+- `containingElement`: The dropdown closes when the user interacts outside of this element.
+- `distance`: Offset distance of the panel away from its trigger.
+- `skidding`: Offset distance of the panel along its trigger.
+- `hoist`: Option to prevent clipping of the panel when placed inside a container with `overflow: auto|scroll`.
+
+## 🔄 Events Update
+
+### Removed Events from `ui-dropdown`:
```
karlbaumhauer, line 50:
```suggestion
### Removed Events:
```

packages/components/src/docs/Migration/ui-dropdown.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+- `entryClick`: Emitted when an option in the menu was clicked.
+- `toggle`: Emitted when the dropdown button was clicked. This event provided the open state of the dropdown.
+
+You can still listen to your slotted `trigger` element or set click listeners on your slotted content.
+
+### New Events in `sd-dropdown`:
```
karlbaumhauer, line 57:
```suggestion
### New Events:
```

packages/components/src/components/dropdown/dropdown.stories.ts
```diff
@@ -0,0 +1,303 @@
...
+  },
+
+  play: async ({ canvasElement }: { canvasElement: HTMLUnknownElement }) => {
+    const trigger = canvasElement.querySelector('.mouseless sd-button');
+    const dropdown = canvasElement.querySelector('.mouseless sd-dropdown');
+    await waitUntil(() => dropdown?.shadowRoot?.querySelector('#dropdown'));
+    await waitUntil(() => trigger?.shadowRoot?.querySelector('button'));
+    await waitUntil(() => dropdown?.shadowRoot?.querySelector('sd-popup:not([active])'));
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    await userEvent.type(trigger!, '{return}', { pointerEventsCheck: 0 });
```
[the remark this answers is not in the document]
mariohamann: Oooooooh yeah! There was a mixture of bugs but this does the focusing much better again like in the button story. Thanks!

karlbaumhauer (changes requested):

packages/components/src/components/dropdown/dropdown.test.ts
```diff
@@ -144,40 +144,44 @@ describe('<sd-dropdown>', () => {
...
-    const trigger = el.querySelector('sd-button')!;
-
-    trigger.focus();
-    await sendKeys({ press: 'ArrowDown' });
-    await el.updateComplete;
-
-    expect(el.open).to.be.true;
-  });
+  /**
+   * TODO: Re-enable this test once sd-menu and sd-menu-item are implemented
```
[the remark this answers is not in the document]
mariohamann: We currently don't even have a `sd-menu` component planned. @coraliefeil Should we do one?
See: https://shoelace.style/components/menu-item

packages/components/src/components/dropdown/dropdown.stories.ts
mariohamann: It is meant for "round" triggers. Therefore this story shows best practice. Showing with a non-round trigger would be wrong. As soon as we have the sd-navigation-item, we can use it instead of the non-rounded-element button.

I'll create a follow-up ticket or add it to the sd-navigation-item ticket!

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
mariohamann: @karlbaumhauer It's fixed now. :)

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
mariohamann: Nah, we can't do that. So the main purpose is to make a good screenshot test here. Usually the dropdown shouldn't open itself but on a user interaction. Then it tries to be visible as good as possible. Even if it's not visible anymore. Now we open all at once to show the behaviour of the positions. What leads to this... ^^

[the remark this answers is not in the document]
mariohamann: If you have better ideas without breaking the behaviour, let me know! :)

packages/components/src/components/dropdown/dropdown.stories.ts
mariohamann: Even after reload?  In fact it's absolutely correct. That's we set `auto-size` to true per default. It resizes itself on the right hand side, but breaks the behaviour on the left-hand side, making the whole page bigger.

It searches for it's best position – and that's at the top, as there is more space available than at the bottom. Indeed I could resize the whole div, so there is more space at the bottom.

mariohamann: I improved the story to have minimum-height in the docs! :) Should look better now as soon as it's fully visible.

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
MarcMatthiae: Not really. I guess we have to leave it like that then.

MarcMatthiae (approved):

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
mariohamann: Did some digging this morning: It works now, at least in the docs! :)

Thanks for requesting changes @karlbaumhauer @MarcMatthiae

packages/components/src/components/dropdown/dropdown.test.ts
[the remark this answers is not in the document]
karlbaumhauer: if we dont have it planned, I'd suggest to remove the comment.

packages/components/src/components/dropdown/dropdown.stories.ts
karlbaumhauer: if its about the rounded triggers, why is the flyout rounded as well? the comment and the shown sample feels confusing.

mariohamann: When triggers are rounded, dropdowns have to be rounded, too.

The example shows a rounded trigger and therefore the `rounded` attribute has to be set to true. Ideas how to improve the docs?

packages/components/src/components/dropdown/dropdown.stories.ts
karlbaumhauer: yes, as soon as it is fully visible. 
While scrolling it still feels very buggy and does not give a good impression of the component... 

can we remove specific stories from the docs and just add a NOTE that this story was excluded due to presentation issues and users should go into the story itself?

mariohamann: Mh – it just shows the correct behaviour, that it looks for a place to show up and people should be aware of that...

...but I would do something differently: I would add a `no-flip` property, so people can opt-out of the automatic flipping behaviour. And we just set this to true in the problematic stories! :)

packages/components/src/components/dropdown/dropdown.stories.ts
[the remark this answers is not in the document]
mariohamann: Fixed for the story as well! :) Thanks for the heads-up

packages/components/src/components/dropdown/dropdown.stories.ts
mariohamann: Added the attribute. Feels absolutely better now. I didn't add no-flip thing to every story, as we still should the default behaviour.

packages/components/src/components/dropdown/dropdown.test.ts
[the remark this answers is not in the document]
mariohamann: Done.

packages/components/src/components/popup/popup.stories.ts
```diff
@@ -1,18 +1,37 @@
...
 
-const { argTypes, args, parameters } = storybookDefaults('sd-popup');
+const { argTypes, parameters } = storybookDefaults('sd-popup');
 const { generateTemplate } = storybookTemplate('sd-popup');
+const { overrideArgs } = storybookHelpers('sd-popup');
 
 /**
  * Popup is a utility that lets you declaratively anchor "popup" containers to another element.
+ *
+ * This component’s name is inspired by <popup>. It uses Floating UI under the hood to provide a well-tested, lightweight, and fully declarative positioning utility for tooltips, dropdowns, and more.
```
karlbaumhauer, line 12:
```suggestion
 * This component’s name is inspired by `<popup>`. It uses Floating UI under the hood to provide a well-tested, lightweight, and fully declarative positioning utility for tooltips, dropdowns, and more.
```

packages/components/src/components/popup/popup.stories.ts
```diff
@@ -1,18 +1,37 @@
...
+ *
+ * Popup doesn’t provide any styles — just positioning! The popup’s preferred placement, distance, and skidding (offset) can be configured using attributes.
+ *
+ * > Popup is a low-level utility built specifically for positioning elements. Do not mistake it for a tooltip or similar because it does not facilitate an accessible experience! Almost every correct usage of <sl-popup> will involve building other components. It should rarely, if ever, occur directly in your HTML.
```
karlbaumhauer, line 16:
```suggestion
 * > Popup is a low-level utility built specifically for positioning elements. Do not mistake it for a tooltip or similar because it does not facilitate an accessible experience! Almost every correct usage of <sd-popup> will involve building other components. It should rarely, if ever, occur directly in your HTML.
```

packages/components/src/components/popup/popup.stories.ts
```diff
@@ -1,18 +1,37 @@
...
  */
 
 export default {
-  title: 'Components/sd-popup',
+  title: 'Utilities/sd-popup',
```
karlbaumhauer, line 20: is this supposed to look like this?
image

packages/components/src/components/dropdown/dropdown.ts
```diff
@@ -468,9 +473,12 @@ export default class SdDropdown extends SolidElement {
         transform-origin: left;
       }
 
-      /* When users slot a menu, make sure it conforms to the popup's auto-size */
-      ::slotted(sd-menu),
-      :host(:not([no-auto-size])) ::slotted(*) {
+      /*
+      * Make sure that slotted content conforms to the popup's auto-size if:
+      * - it's an sd-menu
```
karlbaumhauer, line 478: do we have an sd-menu in the future?

packages/components/src/components/dropdown/dropdown.test.ts
```diff
@@ -144,10 +144,6 @@ describe('<sd-dropdown>', () => {
     expect(panel.hidden).to.be.true;
   });
 
-  /**
-   * TODO: Re-enable this test once sd-menu and sd-menu-item are implemented
-   */
-
   // it('should still open on arrow navigation when no menu items', async () => {
```
karlbaumhauer, line 147: what about those tests now?

karlbaumhauer (changes requested):

karlbaumhauer:
@mariohamann I have 2 additional questions:
1. is it possible to prevent ALL sd-dropdowns in the docs to close as soon as I click on ANY trigger? Afterwards I have to open them ALL manually one-by-one...
2. the dropdown closes as soon as I leave the focus with the keyboard. This is the same behavior as with the mouse but I personally would expect to only close it with escape or hitting enter on the the trigger element again... Not sure about this one.

packages/components/src/components/popup/popup.stories.ts
mariohamann: Yes. This is unstyled on purpose as this is just an unstyled utility component.

karlbaumhauer: I understand but can we move it at least to be completely visible in this story or is it too complicated? looks broken when opening it.

mariohamann:
> @mariohamann I have 2 additional questions:
> 
> 1. is it possible to prevent ALL sd-dropdowns in the docs to close as soon as I click on ANY trigger? Afterwards I have to open them ALL manually one-by-one...
> 2. the dropdown closes as soon as I leave the focus with the keyboard. This is the same behavior as with the mouse but I personally would expect to only close it with escape or hitting enter on the the trigger element again... Not sure about this one.

1. No. Reload or cleaning up the URL helps. This is the same in other interactive components, too.
2. This is in purpose an aligns with a11y behaviour I tested over here: https://www.w3.org/WAI/tutorials/menus/flyout/#use-parent-as-toggle – a different behaviour is used for application menus (that's what the whole `sd-menu` logis is for): https://www.w3.org/WAI/tutorials/menus/application-menus/ where arrow handling and everything is completely different.

packages/components/src/components/dropdown/dropdown.ts
mariohamann: I created a follow-up ticket to either remove logic + tests regarding `sd-menu` or build the component: https://github.com/solid-design-system/figma/issues/111

packages/components/src/components/dropdown/dropdown.test.ts
mariohamann: I created a follow-up ticket to either remove logic + tests regarding `sd-menu` or build the component: https://github.com/solid-design-system/figma/issues/111

packages/components/src/docs/Migration/ui-dropdown.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+- `title` and `icon`: Previously used for the dropdown's appearance.
+- `iconOption`: Determined if an icon can be added.
+- `titleDisplayPosition`: Controlled the title's positioning.
+- `links` and `utilityDropdownLinks`: Managed the links in the dropdown.
+- `downloads` and `untilityDownloads`: Managed download links in the dropdown.
+
+These functionalities are now inherently managed through the default and "trigger" slots in `sd-dropdown`, streamlining the component structure and allowing for more customization.
+
+## 🔀 Updated Attributes
```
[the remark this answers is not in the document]
mariohamann: I added a follow-up ticket:

https://github.com/solid-design-system/solid/issues/357

packages/components/src/components/popup/popup.stories.ts
mariohamann: Ah, got it. Yeah, sure!

karlbaumhauer (approved):

PR #365, opened 2023-09-01
feat: sd-interactive

mariohamann:
## Description:
This PR introduces a new style which is a low-level utility for interactive elements without padding.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/interactive/interactive.declaration.ts
+++ b/packages/components/src/styles/interactive/interactive.declaration.ts
@@ -1,0 +1,22 @@
+import type { Style } from '../../declaration';
+
+export default {
+  styleName: 'sd-interactive',
+  summary: 'Generates basic styles for interactive elements.',
+  status: 'experimental',
+  since: '1.9',
+  attributes: [
+    {
+      name: 'sd-interactive--disabled',
+      description: 'Makes an element look disabled.'
+    },
+    {
+      name: 'sd-interactive--selected',
+      description: 'Makes an element look active.'
+    },
+    {
+      name: 'sd-interactive--reset',
+      description: 'Resets the default browser styles of e. g. a button.'
+    }
+  ]
+} satisfies Style;
--- a/packages/components/src/styles/interactive/interactive.declaration.ts
+++ b/packages/components/src/styles/interactive/interactive.declaration.ts
@@ -4,7 +4,7 @@ export default {
   styleName: 'sd-interactive',
   summary: 'Generates basic styles for interactive elements.',
   status: 'experimental',
-  since: '1.10',
+  since: '1.11',
   attributes: [
     {
       name: 'sd-interactive--disabled',
--- a/packages/components/src/styles/interactive/interactive.declaration.ts
+++ b/packages/components/src/styles/interactive/interactive.declaration.ts
@@ -4,15 +4,11 @@ export default {
   styleName: 'sd-interactive',
   summary: 'Generates basic styles for interactive elements.',
   status: 'experimental',
-  since: '1.9',
+  since: '1.10',
   attributes: [
     {
       name: 'sd-interactive--disabled',
       description: 'Makes an element look disabled.'
-    },
-    {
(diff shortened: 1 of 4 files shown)


mariohamann:
@coraliefeil I'm not happy with the `active` variant. We need this e. g. for the tooltip, but I'm not able to set a proper coloring, as this is the `accent` color which is only available as one single color for text. I think we should add something for hover and active here...

packages/components/src/styles/interactive/interactive.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+  render: (args: any) => {
+    return generateTemplate({
+      options: { templateContent: '<button class="%CLASSES%">%SLOT%</button>' },
+      args
+    });
+  }
+};
+
+/**
+ * Use the `disabled` class to make a interactive element look disabled. This works as well when setting an `disabled` attribute on the element.
```
MarcMatthiae, line 41:
```suggestion
 * Use the `disabled` class to make an interactive element look disabled. This works as well when setting an `disabled` attribute on the element.
```

packages/components/src/styles/interactive/interactive.stories.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+      options: { templateContent: '<button class="%CLASSES%">%SLOT%</button>' },
+      constants: { type: 'attribute', name: 'sd-interactive--disabled', value: true },
+      args
+    });
+  }
+};
+
+/**
+ * Use the `selected` class to make a interactive element look selected.
```
MarcMatthiae, line 59:
```suggestion
 * Use the `selected` class to make an interactive element look selected.
```

MarcMatthiae (approved):
Just two grammatical remarks, rest looks good to me :)

packages/components/src/styles/interactive/interactive.declaration.ts
```diff
@@ -0,0 +1,22 @@
+import type { Style } from '../../declaration';
+
+export default {
+  styleName: 'sd-interactive',
+  summary: 'Generates basic styles for interactive elements.',
+  status: 'experimental',
+  since: '1.9',
```
MarcMatthiae, line 7: Probably needs to be changed before merging, we are currently at 1.9. BTW: What is the next version? 2.0 or 1.10?

karlbaumhauer: @MarcMatthiae as it stays a feature, it will become 1.10. major version bumps (2.0) only apply for breaking changes.

yoezlem:
Don't we have a Figma component for sd-interactive?

karlbaumhauer:
> @coraliefeil Please recheck. I removed "selected".
> 
> --
> 
> Regarding `reset`: This is a little, optional helper for devs to reset default browser styles.
> 
> CleanShot 2023-09-05 at 09 19 22@2x
> We don't want to apply this per default, as projects maybe already have a resetter or set different styles or whatever. :)

@mariohamann wouldnt that be something we could offer as a global Style or do you only see this necessary for the interactive Style?

mariohamann:
> @mariohamann wouldnt that be something we could offer as a global Style or do you only see this necessary for the interactive Style?

Yah, I get the idea. Let's collect that type of occurences of little helpers!

At the moment I would like prevent such utilities as own thing, as styles are already a "new" concept people have to get used to and we have to check out stuff like code splitting etc. If we get a better feeling for everything, I would love to have a second look on that.

karlbaumhauer:
> > @mariohamann wouldnt that be something we could offer as a global Style or do you only see this necessary for the interactive Style?
> 
> Yah, I get the idea. Let's collect that type of occurences of little helpers!
> 
> At the moment I would like prevent such utilities as own thing, as styles are already a "new" concept people have to get used to and we have to check out stuff like code splitting etc. If we get a better feeling for everything, I would love to have a second look on that.

sounds good. Pls create a ticket for the backlog.

PR #432, opened 2023-09-25
chore: fix linting errors on main

mariohamann:
## Description:
Per accident I commited directly to main, which was raising linting bugs. This fix removes that.


yoezlem (approved):
LGTM

Issue #281, opened 2023-07-26
feat: ✨ add sd-video


karlbaumhauer:
## Description
The `sd-video` component can be used to wrap external video elements (e. g. from Moving Image or bare `<video>`-Tags) and provide some basic styling for Union Investment. Developers should be able to wire it up with their video service as they need it (e. g. Moving Image), by setting the attributes they need in combination with events they are listening to.

<img width="700" alt="CleanShot 2023-08-28 at 15 44 54@2x" src="https://github.com/solid-design-system/solid/assets/26542182/6feb526b-31fa-4277-9f3c-775f968ccd9f">

Developers should be able to individually show the poster images (on video element) or overlay. When clicking on the play-button, everything should be hidden (ideally with a transition). Therefore overlay and play button should be in one single container which is hidden then.

The play button is the only element which always has to show up, if `playing` is false, as this is needed at least for a11y.

All logic concerning MovingImage, iFrame handling there etc. should stay in the CMS project. This makes it possible to not be restricted too MovingImage, or `<video>` but even could be used in combination wth other services. We would on top save a bunch JS only used for MovingImage... + the iFrame Service + Tracking Logic for Moving Image etc.

## API
https://www.figma.com/file/q8BLLhesaUN88yF46DGNDI/Video?type=design&node-id=536-7064&mode=design&t=aix2pvBggeAfJ8r5-4
needs Update from https://github.com/solid-design-system/figma/issues/125
## Hand-off
https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=2624-18899&mode=design&t=6KqICmYgdsyxxBIT-4
needs Update from https://github.com/solid-design-system/figma/issues/125
### Components detailed requirements
- [ ] Props, CSS-Properties, Parts & Slots
  - [x] @prop `playing` (hides everything when true)
  - [x] @prop `overlay` (show a darker overlay)
  - [x] @slot `play-icon` (to change the icon)
  - [x] @event `sd-play` (when play button is clicked → this should set `playing` to true and remove all buttons etc.)
  - [ ] @method `play()` (changes everything to start and stop the video and hides the overlay and button)
<del>- [ ] @slot `copyright` (**_not_** implemented as this will be part of the sd-media component)</del>
<del>- [ ] @slot/@prop `poster` => **_not_** implemented as this is part of the video itself (html video tag, MovingImage Video, etc.)</del>
- [ ] Samples
[...]

PR #451, opened 2023-10-09
feat: ✨ sd-video

abudd1094:
adds sd-video component: https://github.com/solid-design-system/solid/issues/281

Please see my comment on the ticket above for status information


packages/components/src/components/video/videoExample.ts
```diff
@@ -0,0 +1,75 @@
...
+            Your browser does not support the video tag.
+          </video>
+          <sd-icon
+            library="global-resources"
+            name="system/start"
+            slot="play-icon"
+            color="primary"
+            style="font-size: 64px; transform: translateX(5px);"
+          ></sd-icon>
+        </sd-video>
```
mariohamann, line 54: This should be part of a sample. This whole file shouldn't exist, as this is not a component.

In the sample We should wire up plain JS with the component.

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,77 @@
...
+        <button
+          aria-label="Play video"
+          tabindex="0"
+          @click=${this.play}
+          class=${cx(
+            this.playing ? 'opacity-0' : 'opacity-100',
+            'absolute left-[50%] top-[50%] translate-x-[-50%] translate-y-[-50%] p-4 bg-white bg-opacity-75 rounded-full flex items-center justify-center transition-opacity duration-300'
+          )}
+        >
+          <slot name="play-icon" part="play-icon"></slot>
```
mariohamann, line 55: The play icon should become a system icon and set as default slot content here.

packages/components/src/components/video/videoExample.ts
```diff
@@ -0,0 +1,75 @@
...
+            <source src="http://media.w3.org/2010/05/sintel/trailer.mp4" type="video/mp4" />
+            Your browser does not support the video tag.
+          </video>
+          <sd-icon
+            library="global-resources"
+            name="system/start"
+            slot="play-icon"
+            color="primary"
+            style="font-size: 64px; transform: translateX(5px);"
+          ></sd-icon>
```
mariohamann, line 53: Please remove.

mariohamann (changes requested):

packages/components/src/components/video/videoExample.ts
abudd1094: is there any other example in the repo to look to?

packages/components/src/components/video/video.ts
abudd1094: What does "become a system icon" mean?

abudd1094: I assume you want:
`<sd-icon library="global-resources" name="system/start" color="primary" class="text-[4rem]"></sd-icon>`

To become something like:
`<sd-icon library="system" name="start" color="primary" class="text-[4rem]"></sd-icon>`

But that doesn't work.  Please provide further instructions

packages/components/src/components/video/videoExample.ts
abudd1094: remove what exactly?  This whole component will be removed as per the comment above

packages/components/src/components/video/video.ts
mariohamann: As soon as you merge main, it should work now as @Vahid1919 just merged an update of https://github.com/solid-design-system/solid/blob/main/packages/components/src/components/icon/library.system.ts in his latest commit to main.

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+import { withActions } from '@storybook/addon-actions/decorator';
+
+const { argTypes, args, parameters } = storybookDefaults('sd-video');
+const { generateTemplate } = storybookTemplate('sd-video');
+
+/**
+ * The `sd-video` component can be used to wrap external video elements in order to provide basic styling for Union Investment.
+ * ## Implementation
+ * Use in combination with a viewer component (e. g. from Moving Image) or a bare `<video>`-Tag in addition to `sd-media`.
+ * `<sd-media>
```
karlbaumhauer, line 14: should we add some information here about the non-existing sd-media component? Or maybe leave it out of here and add it as soon as the component is ready.

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
+import { css, html } from 'lit';
+import { customElement } from '../../../src/internal/register-custom-element';
+import { HasSlotController } from '../../internal/slot';
+import { property } from 'lit/decorators.js';
+import cx from 'classix';
+import SolidElement from '../../internal/solid-element';
+
+/**
+ * @summary Used to wrap external video elements (e. g. from Moving Image or bare <video>-Tags) and provide some basic styling for Union Investment.
+ * @status experimental
```
karlbaumhauer, line 10: pls change before merging

karlbaumhauer (changes requested):

packages/components/src/components/video/video.stories.ts
mariohamann: +1 – would remove everything related `sd-media`

packages/components/src/components/video/video.ts
mariohamann: You mean the `status`?

mariohamann:
@solid-design-system/design 

I know, that the play button is mathematically correct centered but it feels visually off.

CleanShot 2023-10-11 at 09 18 34@2x

I would propose to add a 4px margin to the left to make it visually more pleasing:

CleanShot 2023-10-11 at 09 20 11@2x

Apparently in Figma there already is a little margin to the left in the icon itself:

CleanShot 2023-10-11 at 09 22 55@2x

@Vahid1919 Do you remember, where the icon was coming from?

mariohamann:
The default story should contain an example where you can play around with the knobs. We should add a placeholder image in format 16:9 for that in the background. 

I just opened a PR for the image: https://github.com/solid-design-system/solid/pull/457

van-nguyen-ht:
@mariohamann yes indeed, in [figma](https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=1663-18047&mode=design&t=YPSuBz0TT42I9sPK-11) the icon has a padding to the left

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * `sd-video` does not have any content by default because it is a wrapper component.
+ */
+
+export const Default = {
+  render: (args: any) => {
+    return generateTemplate({ args });
+  }
+};
```
mariohamann, line 37: As mentioned here https://github.com/solid-design-system/solid/pull/451#issuecomment-1757026290 we should show at least an image in the main slot.

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
...
+  /** True if the contained video is playing. Hides everything when true. */
+  @property({ type: Boolean, reflect: true }) playing = false;
+
+  /** Show a darker overlay. */
+  @property({ type: Boolean, reflect: true }) overlay = false;
+
+  play() {
+    this.emit('sd-play');
+    this.playing = true;
+  }
```
mariohamann, line 36: If we provide this, we should have a second method, that does the opposite.

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @summary Used to wrap external video elements (e. g. from Moving Image or bare <video>-Tags) and provide some basic styling for Union Investment.
+ * @status experimental
+ * @since 1.0
+ *
+ * @dependency sd-example
+ *
+ * @event sd-play - Event emitted when the video is played or paused. Detail reflects "playing" property.
+ *
+ * @slot - The default slot used to pass a video player element.
+ * @slot play-icon - The video's play icon.
```
mariohamann, line 18: ❤️

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart play-button - Button element wrapper around the play-icon slot.
+ */
+@customElement('sd-video')
+export default class SdVideo extends SolidElement {
+  private readonly hasSlotController = new HasSlotController(this, '[default]', 'play-icon');
+
+  /** True if the contained video is playing. Hides everything when true. */
```
mariohamann, line 27:
```suggestion
  /** Set to `true` to hide the play icon and the overlay. */
```

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
...
+  @property({ type: Boolean, reflect: true }) playing = false;
+
+  /** Show a darker overlay. */
```
mariohamann, line 30:
```suggestion
  /** Set to `true` to show a dark overlay. Only used when `playing` is `false`. */
```

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,86 @@
...
+          )}
+        ></div>
+        <button
+          part="play-button"
+          aria-label="Play video"
+          tabindex="0"
+          @click=${this.play}
+          class=${cx(
+            this.playing ? 'opacity-0' : 'opacity-100',
+            'absolute left-[50%] top-[50%] translate-x-[-50%] translate-y-[-50%] p-4 bg-white bg-opacity-75 rounded-full flex items-center justify-center transition-opacity duration-300'
```
mariohamann, line 57: "transition" === 😍

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+
```
mariohamann, line 38: When implementing the placeholder, make another story which does `Playing x Overlay`.

mariohamann (changes requested):

packages/components/src/components/video/video.ts
abudd1094: what should status be?

karlbaumhauer: status and version.

Status should be stable and version needs to be updated to the latest +1 right before merging.

packages/components/src/components/video/video.ts
abudd1094: I don't think it makes sense to expose this or add a "pause" method, control is always coming from the wrapped component and should be reflected via the property.  

I propose marking this function as private and using only for the click event to make it simple.  External control should use the "playing" property exclusively.

abudd1094:
> @solid-design-system/design
> 
> I know, that the play button is mathematically correct centered but it feels visually off.
> 
> CleanShot 2023-10-11 at 09 18 34@2x
> I would propose to add a 4px margin to the left to make it visually more pleasing:
> 
> CleanShot 2023-10-11 at 09 20 11@2x
> Apparently in Figma there already is a little margin to the left in the icon itself:
> 
> CleanShot 2023-10-11 at 09 22 55@2x
> @Vahid1919 Do you remember, where the icon was coming from?

I had a translate on there I thought you said to "remove".  Margin will distort the background:
image

I will re-add the translate

packages/components/src/components/video/video.ts
karlbaumhauer: good point... @mariohamann what do you think?

packages/components/src/components/video/video.ts
```diff
@@ -7,12 +7,12 @@ import SolidElement from '../../internal/solid-element';
 
 /**
  * @summary Used to wrap external video elements (e. g. from Moving Image or bare <video>-Tags) and provide some basic styling for Union Investment.
- * @status experimental
- * @since 1.0
+ * @status stable
+ * @since 1.16.0
```
karlbaumhauer, line 11:
```suggestion
 * @since 1.19.0
```

karlbaumhauer (changes requested):

Vahid1919:
> @Vahid1919 Do you remember, where the icon was coming from?

It was from the icon-library from the previous Union Investment project I believe.

packages/components/src/components/video/video.ts
mariohamann: Ok. Either both or none. In accordion we provide it like this:

CleanShot 2023-10-11 at 13 23 20@2x

But I think it's totally valid to just work with the prop here.

mariohamann:
> > @Vahid1919 Do you remember, where the icon was coming from?
> 
> It was from the icon-library from the previous Union Investment project I believe.

It seems like the icon is not correctly set in it's frame. Could you please add a ticket, that we have to include its padding and remove the added translate by Alec then?

mariohamann:
I could get the video playing when doing it via keyboard.

https://64469bd84c5ea1ce2dd5d016-nziojohwnh.chromatic.com/?path=/story/components-sd-video--video-element

mariohamann:
So we should have one default story with jsut the imag e + one Overlay x Playing Story. @abudd1094  :)

Vahid1919:
I noticed that the default play button is still visible underneath our own for Firefox --> 
image

PS: This only seems to be the case for the first button press, it does not come back afterwards.

abudd1094:
> I could get the video playing when doing it via keyboard.
> 
> https://64469bd84c5ea1ce2dd5d016-nziojohwnh.chromatic.com/?path=/story/components-sd-video--video-element
> 
> And I'm not sure regarding the focus styles – I think we should use our default ones.

By "default" you mean: focus-visible:focus-outline

mariohamann:
@karlbaumhauer aybe we should reintroduce the `poster` slot – this would prevent everything here?

karlbaumhauer:
> @karlbaumhauer aybe we should reintroduce the `poster` slot – this would prevent everything here as it could just be an overlay over the video?

thats a good idea...

abudd1094:
@karlbaumhauer 

- I removed my translate for the icon as requested, Vahid's PR should fix that without further changes in the sd-video.
- The play icon is centered in our component, which does not take into account the "controls" in the underlying video component.  

I'll add a poster slot, we'll have a quick check in the daily, and I'll add stories / tests if that is accepted.

mariohamann:
@abudd1094 @karlbaumhauer  An alternative could be to only show controls when playing. This could work when you extend your JS, wouldn't it?

abudd1094:
@karlbaumhauer @mariohamann 

OK, after much pain and learning, sd-video is ready for another review.

I have added a "poster" slot which seeks to mirror the behavior of the "poster" attribute on the native "video" element.  It will be shown under the optional overlay and on top of the default video element slot with no pointer events.  So the user must click (or use keyboard to activate) the "play-icon" for the initial play, after which the "poster" will be hidden, and the user can click through the overlay to the video controls.  

I have added as many tests as I could think to add.  I would also like to add the basic "accessibility" test, but I'm not sure I understand how it works.  It seems to fail for Chrome here but I cannot figure out the error:
image

Accessibility has been further bolstered and I added a story for "mouseless" control.  Please advise.

karlbaumhauer:
@abudd1094 sorry for the delay... I will have a look into it by tomorrow.

packages/components/src/components/video/video.stories.ts
```diff
@@ -21,21 +36,109 @@ export default {
...
         args
       })}
     `;
   }
 };
 
+// Script used in next 2 stories
+const videoElementScript = html`<script>
+  function runScriptOnPathChange() {
+    // *NOTE: Storybook single page behavior complicates queries between the "Docs" and "Video Element" stories.
```
karlbaumhauer, line 84: I like that one...!

packages/components/src/components/video/video.stories.ts
```diff
@@ -21,21 +36,109 @@ export default {
...
+  runScriptOnPathChange();
+
+  // Run the script again whenever the hash (path) changes
+  window.onhashchange = runScriptOnPathChange;
+</script>`;
+
 /**
- * Use in combination with a viewer component (e. g. from Moving Image) or a bare `<video>`-Tag. Here we wrap a `video` element that includes a `poster` attribute.
+ * Use in combination with a viewer component (e. g. from Moving Image) or a bare `<video>`-Tag. <br/>
+ * Here we wrap a `video` element that includes a `poster` attribute.
```
karlbaumhauer, line 141:
```suggestion
 * Here we wrap a `video` element that includes setting an image to the `poster` slot.
```

packages/components/src/components/video/video.stories.ts
```diff
@@ -47,31 +150,45 @@ export const VideoElement = {
...
+          constants: [videoConstant, posterConstant]
+        })}
+        ${videoElementScript}
+      </div>
     `;
   }
 };
+
+/**
+ * Duplicates the previous story with default focus on the play button.
```
karlbaumhauer, line 165: we need to deactivate the mouse in this story, when hovering over the component. Have a look into the sd-tag mouseless story to see how its implemented.

karlbaumhauer (changes requested):

karlbaumhauer:
> @karlbaumhauer @mariohamann
> 
> OK, sd-video is finally ready for another review.
> 
> I have added a "poster" slot which seeks to mirror the behavior of the "poster" attribute on the native "video" element. It will be shown under the optional overlay and on top of the default video element slot with no pointer events. So the user must click (or use keyboard to activate) the "play-icon" for the initial play, after which the "poster" will be hidden, and the user can click through the overlay to the video controls.
> 
> I have added as many tests as I could think to add. I would also like to add the basic "accessibility" test, but I'm not sure I understand how it works. It seems to fail for Chrome here but I cannot figure out the error:
> 
> Accessibility has been further bolstered and I added a story for "mouseless" control. Please advise.

So the component looks good and I just found some small text changes and 1 issue in the mouseless story.

Regarding the tests, I unfortunately cannot help here. @Vahid1919 or @mariohamann pls advise if possible.

mariohamann:
CleanShot 2023-10-18 at 10 18 00@2x

Could you please set the size on the image to prevent this?

I think you're good to go with `class="w-WHATEVER-WIDTH aspect-video cover"`.

mariohamann:
Please use one of our internal images as poster. As these will be used for screenshot tests, we shouldn't have any external dependencies on them. (That's why we e. g. intercept all calls to global-resources and provide own icons, as we don't want to have any external stuff.) 

CleanShot 2023-10-18 at 10 20 47@2x

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,194 @@
...
+import type { ConstantDefinition } from '../../../scripts/storybook/helper';
+
+const { argTypes, args, parameters } = storybookDefaults('sd-video');
+const { generateTemplate } = storybookTemplate('sd-video');
+
+const videoConstant: ConstantDefinition = {
+  type: 'slot',
+  name: 'default',
+  value:
+    '<video controls id="videoExample" style="min-width: 854px; min-height: 480px; width: 854px; height: 480px;"><source src="http://media.w3.org/2010/05/sintel/trailer.mp4" type="video/mp4" />Your browser does not support the video tag.</video>'
```
mariohamann, line 16: You can use here Tailwind as well. Something like `w-ANYWIDTH aspect-video cover`.

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    function updatePlayingStatus() {
+      sdVideo.playing = videoEl.paused ? false : true;
+    }
+
+    function removeEventListeners() {
+      // ensure we affect ANY existing elements by querying for all, see *NOTE above
+      const videoEls = document.querySelectorAll('#videoExample');
+      const sdVideos = document.querySelectorAll('#sdVideoExample > sd-video');
+      const sdVideoPlayButtons = document.querySelectorAll('#sdVideoExample > sd-video');
```
mariohamann, line 112: Could you add some more documentation for the users of our library?

Pointing them to what's important for them etc. Some stuff isn't important, like Storybook changes, maybe you can clarify that, too.

So more thinking from a user of our library: What is importatnt. What do I have to do why. What's not important.

packages/components/src/components/video/video.stories.ts
```diff
@@ -0,0 +1,194 @@
...
+export const VideoElement = {
+  name: 'Video Element',
+  parameters: {
+    controls: {
+      exclude: ['playing', 'default']
+    }
+  },
+  render: (args: any) => {
+    return html`
+      <div id="sdVideoExample">
```
mariohamann, line 153:
```suggestion
      <div id="sd-video-example">
```

packages/components/src/components/video/video.test.ts
```diff
@@ -0,0 +1,137 @@
...
+import type SdVideo from './video';
+
+const defaultSlot = html`<video controls>
+  <source src="http://media.w3.org/2010/05/sintel/trailer.mp4" type="video/mp4" />
+  Your browser does not support the video tag.
+</video>`;
+const posterSlot = html`<img
+  slot="poster"
+  alt="poster"
+  src="https://www.blender.org/wp-content/uploads/2020/10/robin-tran-redautumnforest_pr1.jpg"
```
mariohamann, line 12: Especially here: Definitely remove external dependencies. If needed just inline the image with base64 or whatever.

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,137 @@
...
+        ${this.hasSlotController.test('poster')
+          ? html`<slot name="poster" part="poster" role="presentation" @transitionend=${this.hidePoster}></slot>`
+          : null}
+        <div
+          part="overlay"
+          id="overlay"
+          role="presentation"
+          class=${cx(
+            this.overlay && !this.playing ? 'opacity-100' : 'opacity-0',
+            'bg-[rgba(0,0,0,0.65)] w-full h-full absolute top-0 left-0 pointer-events-none z-20 play-pause-transition'
```
mariohamann, line 85:
```suggestion
            'bg-black/65 w-full h-full absolute top-0 left-0 pointer-events-none z-20 play-pause-transition'
```

Couldn't you use something like bg-black/65 or something?

packages/components/src/components/video/video.ts
```diff
@@ -0,0 +1,137 @@
...
+        ></div>
+        <button
+          part="play-button"
+          aria-label="Play video"
+          tabindex="0"
+          @click=${this.play}
+          @keydown=${this.play}
+          class=${cx(
+            this.playing ? 'opacity-0 pointer-events-none' : 'opacity-100',
+            'absolute left-[50%] top-[50%] translate-x-[-50%] translate-y-[-50%] p-4 bg-white bg-opacity-75 rounded-full flex items-center justify-center z-30 play-pause-transition'
```
mariohamann, line 96:
```suggestion
            'absolute left-1/2 top-1/2 translate-x-1/2 translate-y-1/2 p-4 bg-white bg-opacity-75 rounded-full flex items-center justify-center z-30 play-pause-transition'
```

Something like this should work.

packages/components/src/components/video/video.stories.ts
abudd1094: You mean via code comments in this block?

packages/components/src/components/video/video.ts
abudd1094: Before:
image

After:
image

packages/components/src/components/video/video.ts
abudd1094: discussed via phone --- no bg-black token exists :(

packages/components/src/components/video/video.stories.ts
mariohamann: Yes

mariohamann:
> > Please use one of our internal images from our placeholders package as poster. As these will be used for screenshot tests, we shouldn't have any external dependencies on them. (That's why we e. g. intercept all calls to global-resources and provide own icons, as we don't want to have any external stuff.)
> > CleanShot 2023-10-18 at 10 20 47@2x
> 
> What about for the video?

As this doesn't show up in the screenshots, I'm less opinionated there. We maybe should have a default there, too. But for the moment I think that's fine.

packages/components/src/components/video/video.test.ts
abudd1094: "just inline the image with base64 or whatever" --- please expand on what you mean.  I don't think you want:
image
I have removed the src completely, it still passes the tests.  The video remains in the tests and is an external dependency.  If you have a better way to do it, please explain, otherwise I'll leave it like this

abudd1094:
All comments have been addressed.  @mariohamann I added this issue regarding the accessibility test:
https://github.com/solid-design-system/solid/issues/483

Please let me know if I should refine this ticket or make further changes.

packages/components/src/components/video/video.stories.ts
```diff
@@ -13,13 +13,18 @@ const videoConstant: ConstantDefinition = {
...
+  type: 'slot',
+  name: 'default',
+  value: '<img class="w-[400px] aspect-video object-cover" src="./placeholders/generic.jpg" />'
 };
 const posterConstant: ConstantDefinition = {
   type: 'slot',
   name: 'poster',
   value:
-    '<img slot="poster" alt="poster" src="https://www.blender.org/wp-content/uploads/2020/10/robin-tran-redautumnforest_pr1.jpg" />'
+    '<img slot="poster" alt="poster" class="w-WHATEVER-WIDTH aspect-video cover" src="./placeholders/architecture.jpg" />'
```
karlbaumhauer, line 27: @mariohamann was this `w-WHATEVER-WIDTH` really what you expected? If yes, why? Would love to then see a description here.

karlbaumhauer (approved):

karlbaumhauer:
> All comments have been addressed. @mariohamann I added this issue regarding the accessibility test: #483
> 
> Please let me know if I should refine this ticket or make further changes.

@abudd1094 I looked at the ticket and re-ordered it a bit. thx for creating it.

abudd1094:
@van-nguyen-ht 
image

Point 1:
It is pixel for pixel the same, 96 x 96, please check the measurements by number:
image
image

Point 2:
"There's no interactions with the play button" --- yes, that is expected.  In the default story, the component uses a placeholder image as its "default slot" so there is no video to play.  Please see the Video Element stories for interaction demonstrations.

coraliefeil:
SRY messed up assignees by accident

abudd1094:
@van-nguyen-ht requested that I implement the queries from the design:
image

@mariohamann I didn't see 414px as a standard breakpoint, it is implemented in the css block.  If there is a more desirable way to implement this or you want to adjust the format please advise.

EllaELSA:
Not sure where to add a comment, here or in chromatic? 
Shouldn't we provide a hover effect on video or at least at the play button?

coraliefeil:
Removed myself ... Van’s taking care of design review

coraliefeil:
@abudd1094 could we use the sd-interactive for the play icon and the close icon? states are the same.

abudd1094:
@EllaELSA 
This appears to be a different element entirely <sd-play-button> which is not mentioned in the ticket.

@coraliefeil 
What is sd-interactive?  I don't see it in Storybook or Figma... I do see a ticket in Review, perhaps it is not yet merged then.  In that case, we can use it, but I'll have to wait until it is merged.  

@mariohamann @karlbaumhauer @yoezlem 
Seems like we missed a dependency.  Should I add <sd-play-button> and include it in the default slot as part of this PR?

Sounds like what we want, is to merge sd-interactive, use it to build the sd-play-button, and then use that as the default slot for sd-video???

That is fine for me but this information was not transparent.  Please advise before I continue.

coraliefeil:
@abudd1094 https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/styles-sd-interactive--docs

mariohamann:
I'm not sure if the `sd-interactive` class already works inside components. Maybe it does, maybe not. I would suggest du create a follow-up ticket for a possible refactoring, but as long as the current styles are fine, we should proceed.

mariohamann:
@abudd1094 You can check an implemetnation example in sd-carousel:

import InteractiveStyles from '../../styles/interactive/interactive.css?inline';

You can sk @Vahid1919 for further implementation details.

mariohamann:
@abudd1094 You could either set `poster` and `overlay` to pointer-events-none, so that clicking will be passed to the element or you set the click handler on the whole element.

@EllaELSA is right, that clicking anywhere should trigger playing: https://www.w3schools.com/html/html5_video.asp https://www.w3schools.com/tags/att_video_poster.asp

Besides that: Hover styles aren't defined anywhere for cover elements (as this isn't defined in design) and doesn't work in current UI videos as well. I suggest this to be an improvement if design catched up with everything.

mariohamann:
@abudd1094 I don't understand the purpose of the `sd-play-button` as well, especially as should never handle pause with a big filling icon on the video?

I would definitely say to skip this component here and we will validate, if there's a reason for the `sd-play-button`. If yes, we might refactor at a later point, but for the moment, I don't see anything we can benefit from it.

abudd1094:
@mariohamann @karlbaumhauer @EllaELSA @coraliefeil @van-nguyen-ht 
Play button now contains "sd-interactive", covers the whole element to field all click events and applies hover style to the play icon.  Click events have been fine tuned for precision.  I added a debounce functionality in the story to prevent flickering of our play icon on "seeking" events, without this, the icon will show up briefly every time a user initiates seeking in the video player.

abudd1094:
77 comments, I'm going for a record...

@van-nguyen-ht @mariohamann 
I moved the media query from screen to element as mentioned on Figma.  This should cover all of the feedback we have gathered here.  Will update PR again when Quality Gate is fixed.

packages/components/src/components/navigation-item/navigation-item.ts
```diff
@@ -10,8 +10,8 @@ import SolidElement from '../../internal/solid-element';
 
 /**
  * @summary Flexible button / link component that can be used to quickly build navigations. Takes one of 3 forms: link (overrides all other if 'href' is provided), button (default), or accordion (if 'children' slot defined).
- * @status experimental
- * @since 1.12.0
+ * @status stable
+ * @since 1.15.0
```
mariohamann, line 14: Is this intended?

mariohamann (approved):

packages/components/src/components/navigation-item/navigation-item.ts
abudd1094: yes, the version was incorrect --- regarding status, it is not clear to me how you determine that, imo this is now stable and not experimental

karlbaumhauer: everything is stable unless we specifically announce something as experimental

abudd1094:
@van-nguyen-ht @coraliefeil @EllaELSA 
I have added you as assignees for the UI Tests at your lesiure.

@Vahid1919 @yoezlem @DanielHargesheimer 
I have adde you as assignees simply to have a look at this new feature (as discussed this morning in the daily).  No need to review!

Issue #450, opened 2023-10-09
fix: 🤔 Tailwind classes not applied to part="content" in Accordion component


yoezlem:
## Current behavior
Currently, it's not possible to modify the paddings within the 'content' part because the spacing is set within the child element below the part. 

## Expected behavior
The block px-4 py-6 classes need to be extracted from the slot and applied to the 'content' part."

## Steps to reproduce
<img width="1076" alt="image" src="https://github.com/solid-design-system/solid/assets/98025014/704bd004-6cac-4979-b80f-a44dcd56b2ce">

## DoR
- [X] Item has business value
- [X] Item has been estimated by the team
- [X] Item is clear and well-defined
- [X] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #454, opened 2023-10-10
fix: 🤔 added part selector to accordion content slot.

yoezlem:
The requirement: https://github.com/solid-design-system/solid/issues/450

### DIFF
--- a/packages/components/src/components/accordion/accordion.stories.ts
+++ b/packages/components/src/components/accordion/accordion.stories.ts
@@ -130,7 +130,7 @@ export const Slots = {
 
 export const Parts = {
   parameters: {
-    controls: { exclude: ['base', 'header', 'summary', 'summary-icon', 'content'] }
+    controls: { exclude: ['base', 'header', 'summary', 'summary-icon', 'content', 'content__slot'] }
   },
   render: (args: any) => {
     return generateTemplate({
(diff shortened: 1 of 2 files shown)


karlbaumhauer (approved):

Vahid1919:
@yoezlem I noticed that the animation is no longer smooth after moving the padding to the parent. It disappears before the closing animation making it stutter on my device.

https://github.com/solid-design-system/solid/assets/63849626/3f2b2592-0289-41e1-841e-c4879464113a


 One possible solution I found is to speed up the animation (duration 150 instead of 300) and that masks this behavior. I couldn't find a better solution personally, but would love to hear other ideas.

mariohamann:
Maybe an alternative approach: Could we give the slot a part selector, too?

yoezlem:
I don't know if its clear enough whats the different between those two is:
image

Maybe we could add here content-container or something like that?
image

Vahid1919:
> Good eye, @Vahid1919.
> 
> Maybe an alternative approach: Could we give the slot a part selector, too?
> 
> `part="content__slot"`

@mariohamann That's a better approach. 

I added an offset for the outlines for the Parts story. That's because "content__slot" is not visible there since the parent has an "overflow-hidden". I added the new changes.

Vahid1919:
> I don't know if its clear enough whats the different between those two is
> Maybe we could add here content-container or something like that? 

Yes, I think adding wording like container makes the most sense

karlbaumhauer:
@Vahid1919 / @yoezlem is it on purpose that this is still a draft?

Vahid1919:
@solid-design-system/development Please Review :) 

@van-nguyen-ht I added you as an assignee regarding the UI Reviews :)

mariohamann (approved):

azraefendic (approved):

PR #456, opened 2023-10-11
docs: ✨ accessibility checklist

azraefendic:
Adds an accessibility checklist to the storybook documentation in order to provide a guidance for designers and developers on how to use our library.

closes https://github.com/solid-design-system/internal/issues/33


packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+# Accessibility
+
+Union Investment strives to offer digital products and services that are accessible to all, regardless of their abilities/limitations or the technology used. This is an accessibility checklist used by the SDS team when designing and developing new features. It helps us improve the experience for everyone who uses our product and ensures our high standards are met.
+
+This is an ongoing process. We will continue to improve accessibility and meet at least the **WCAG 2.1 Level AA guidelines** - also with a view to the Barrier-Free Accessibility Strengthening Act that will come into force from 2025.
+
+## Developing a WCAG-compliant Basis
+
+We are committed to creating products that conform to the WCAG 2.1 guidelines, a standard also embraced by AK CD. This commitment includes a range of specific practices:
```
karlbaumhauer, line 13:
```suggestion
We are committed to creating products that are conform to the WCAG 2.1 guidelines, a standard also embraced by the Union Investment brand department. This commitment includes a range of specific practices:
```

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+- **Preserving ARIA Information**: Do not delete the ARIA information from the solid examples. They ensure correct interpretation by screen readers.
+
+### Multimedia:
+
+- **Alt text**: Provide alternative text for images, icons, and non-text content.
+
+- **Captions and transcripts**: Always include captions and transcripts for videos and audio content.
+
+- **Avoid auto-playing** audio and video without user controls.
```
karlbaumhauer, line 33: I am not sure about this one. AFAIK (but pretty old information), autoplay is not allowed. Firefox even banned it for a while from the browser.

If so, we should rephrase this to something like "Do not use auto-playing"

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+### Color:
+
+- **Color contrast**: Verify that color contrast meets the WCAG AA standard.
```
karlbaumhauer, line 37:
```suggestion
- **Color contrast**: [Verify that color contrast](https://color.a11y.com/) meets the WCAG AA standard.
```

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+- **Mouseless stories** to test and ensure all functionality is accessible without relying on mouse interactions.
+
+- **A11y test cases**.
+
+- **Focus visibility and order** for interactive elements.
+
+## Integration into Processes as Definition-of-Done (DOD)
+
+Accessibility is not an afterthought. It is integrated into our **Definition-of-Done**. This means that no work is considered complete until it meets our accessibility standards.
```
karlbaumhauer, line 50: ❤️

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## The Importance of Web Accessibility
+
+By making your website accessible, you ensure that all users, including people with disabilities, have a good user experience and can easily access information. Web accessibility includes all disabilities that affect access to the Internet, including auditory, cognitive, neurological, physical, speech, visual, and age-related impairments.
+
+In Germany, about 0.4 percent of the population is visually impaired and about 0.1 percent of the inhabitants are auditorily impaired. Making sure our products are accessible opens them up to a wider audience and ensures that we are not excluding any potential users.
```
karlbaumhauer, line 56: where does this come from? 
Just asking as the 0.5 percent does not really boost the user numbers. I would rather go with the fact that nobody should be excluded and especially as a financial service provider, we cannot withheld this service to some people...

But maybe only a personal point of view.

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## What's New
```
karlbaumhauer, line 58:
```suggestion
## What's New in the Solid Design System
```

karlbaumhauer: or maybe some other phrasing as it should be clear whats new compared to what... basically this is all new compared to the time before we had the SDS, right?

packages/components/src/docs/General/Accessibility.mdx
```diff
@@ -0,0 +1,74 @@
+import { Meta } from '@storybook/blocks';
```
karlbaumhauer, line 1: the file needs to be renamed into "7-Accessibility" right @mariohamann?

I would even rename it to "6-Accessibility" and rename the browser issue to "7-Known Browser Issues"

karlbaumhauer (changes requested):

packages/components/src/docs/General/Accessibility.mdx
karlbaumhauer: pls also share this with the design team and maybe we need a follow up ticket to align it.

karlbaumhauer (approved):

yoezlem (approved):

packages/components/src/docs/General/Accessibility.mdx
azraefendic: Will do!

Vahid1919 (approved):

coraliefeil:
I would adjust:

Multimedia
Never auto-play audio and video.

to

Multimedia
Never auto-play audio and video without providing pause/play functionalities.

--

I would add:

Motion:
Include CSS prefers-reduced-motion for all web content.

--

Thanks

abudd1094 (approved):

EllaELSA:
Thanks for the list. 
Just one thing... more a question what about pointing out responsiveness. Should we mention that this is also an area for accessibilty? As we all know that not everything in the UI universe is actually usable in mobile.

karlbaumhauer:
> Thanks for the list. Just one thing... more a question what about pointing out responsiveness. Should we mention that this is also an area for accessibilty? As we all know that not everything in the UI universe is actually usable in mobile. @coraliefeil @yoezlem @karlbaumhauer @mariohamann

like the idea very much. @azraefendic maybe you can add something to the checklist?!?

karlbaumhauer (approved):

DanielHargesheimer (approved):

mariohamann (approved):

PR #507, opened 2023-10-30
feat: ✨ sd-checkbox & sd-checkbox-group

DanielHargesheimer:
As discussed in the team, there is no invalid prop and the error styles depend on the data-user-invalid attribute shoelace provides. Because of this it is not possible to show the error styles in storybook. You can test it in the required story: Click on a checkbox, click again to uncheck it and leave it, then the error styles will be shown.


mariohamann:
@DanielHargesheimer Could you create an interactive Story for the error state?

packages/components/src/components/checkbox-group/checkbox-group.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+  args: overrideArgs([
+    { type: 'slot', name: 'label', value: `<label slot="label">Group Label</label>` },
+    {
+      type: 'slot',
+      name: 'default',
+      value: `<sd-checkbox name="checkbox" value="1">Checkbox 1</sd-checkbox><sd-checkbox name="checkbox" value="2">Checkbox 2</sd-checkbox><sd-checkbox name="checkbox" value="3">Checkbox 3</sd-checkbox>`
+    }
+  ]),
+  argTypes,
+  parameters: { ...parameters }
```
yoezlem, line 23:
```
  parameters: { 
      ...parameters,
      design: {
          type: 'figma',
          url: 'https://www.figma.com/file/Q7E9GTBET7Gs2HyH1kbpu5/Checkbox-%2F-Checkbox-Group?type=design&node-id=0-1&mode=design&t=DV2yJRUqqYBrskyb-0'
    }
  }
```

Don't forget to add the design link in your story as reference! 🙂

packages/components/src/components/checkbox-group/checkbox-group.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+        x: { type: 'attribute', name: 'orientation' },
+        y: { type: 'attribute', name: 'size' }
+      },
+      args
+    });
+  }
+};
+
+/**
+ * Use the disabled attribute to disable a input checkbox. Clicks will be suppressed until the disabled state is removed
```
yoezlem, line 54:
```suggestion
 * Use the disabled attribute to disable an input checkbox. Clicks will be suppressed until the disabled state is removed
```

yoezlem:
I am missing the story for the 'invalid' and 'required' states. Additionally, we also need a sample story for the 'indeterminate' state.

packages/components/src/components/checkbox/checkbox.ts
```diff
@@ -185,35 +197,96 @@ export default class SdCheckbox extends SolidElement implements SolidFormControl
...
+                  name="status-minus"
                 ></sd-icon>
               `
-        : ''}
+            : ''}
+        </span>
+        <span
+          part="label"
+          class=${cx(
+            'checkbox__label select-none inline-block ml-2 text-[var(--sd-input-label-color)]',
```
yoezlem, line 243: text-[var(--sd-input-label-color)] please change this to text-black

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  private async syncCheckboxElements() {
+    const checkboxes = this.getAllCheckboxes();
+
+    await Promise.all(
+      // Sync the checked state and size
+      checkboxes.map(async checkbox => {
+        await checkbox.updateComplete;
+
+        checkbox.size = this.size;
```
yoezlem, line 56: I din't fully understand how you handle the invalid state from sd-checkbox to sd-checkbox-group.

Why haven't you created an 'invalid' property in 'sd-checkbox' and synced it like checkbox.size?

DanielHargesheimer:
> I am missing the story for the 'invalid' and 'required' states. Additionally, we also need a sample story for the 'indeterminate' state.

For the checkbox I've added the interactive invalid story. There is a required story for the checkbox. 
The checkbox-group cannot be required, there is no validation on the checkbox-group. 

The indeterminate state is displayed in a checkbox story. We don't provide any logic in the checkbox-group to show the indeterminate state in a checkbox group. If we want to do that, we have to discuss it.

yoezlem:
What about these two states?
image

yoezlem:
And also these:
image

yoezlem:
The icon looks a little bit weird..? Is this the right one? Why is it so pixelated?
image

DanielHargesheimer:
> What about these two states? image

Right now, the checked one is impossible. Checked is always valid. The indeterminate state I can show, you're right.

yoezlem:
Why can you only use the Tab key when using 'keyboardless'? Wouldn't it be sensible to also be able to use arrow keys like (ArrowUp, ArrowDown, ArrowLeft, ArrowRight)?

DanielHargesheimer:
> Why can you only use the Tab key when using 'keyboardless'? Wouldn't it be sensible to also be able to use arrow keys like (ArrowUp, ArrowDown, ArrowLeft, ArrowRight)?

This is not the common behavior on multiple checkboxes in a fieldset. It's different from the radio-group.

yoezlem:
> > I am missing the story for the 'invalid' and 'required' states. Additionally, we also need a sample story for the 'indeterminate' state.
> 
> For the checkbox I've added the interactive invalid story. There is a required story for the checkbox. The checkbox-group cannot be required, there is no validation on the checkbox-group.
> 
> The indeterminate state is displayed in a checkbox story. We don't provide any logic in the checkbox-group to show the indeterminate state in a checkbox group. If we want to do that, we have to discuss it.

@mariohamann @coraliefeil Is this how the checkbox/checkbox-group is supposed to behave?

packages/components/src/components/checkbox-group/checkbox-group.ts
DanielHargesheimer: Because the checkbox is a FormControl and has it's own validation. This is the main difference to the radio component. The radio component needs the radio-group component to pass in the invalid state.
We cannot do this for the checkbox-group and checkbox as we don't know what's valid and invalid in a checkbox-group.

mariohamann:
> > > I am missing the story for the 'invalid' and 'required' states. Additionally, we also need a sample story for the 'indeterminate' state.
> > 
> > 
> > For the checkbox I've added the interactive invalid story. There is a required story for the checkbox. The checkbox-group cannot be required, there is no validation on the checkbox-group.
> > The indeterminate state is displayed in a checkbox story. We don't provide any logic in the checkbox-group to show the indeterminate state in a checkbox group. If we want to do that, we have to discuss it.
> 
> @mariohamann @coraliefeil Is this how the checkbox/checkbox-group is supposed to behave?

I don't see another way. Activating the indeterminate state can only be done by the developer's in an application, see e. g.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

If a checkbox has sub-checkboxes, they have to track, whether those sub-checkboxes are all checked or not.

mariohamann:
> > Why can you only use the Tab key when using 'keyboardless'? Wouldn't it be sensible to also be able to use arrow keys like (ArrowUp, ArrowDown, ArrowLeft, ArrowRight)?
> 
> This is not the common behavior on multiple checkboxes in a fieldset. It's different from the radio-group.

Correct.

packages/components/src/components/checkbox-group/checkbox-group.ts
mariohamann: What @DanielHargesheimer says. The checkbox can and has to check itself. The check group can't know anything about validation, as it could be everything (check all cars in a list, check a max of 2, check a min of 2 etc.) and can only be handled by the application defining a validation.

DanielHargesheimer:
> The icon looks a little bit weird..? Is this the right one? Why is it so pixelated? image

I don't know, I've added the svg from the figma file.

coraliefeil:
Approved. THX! Good job.

yoezlem (approved):

azraefendic:
Looks good, but are we missing the migration guide here?

packages/components/src/components/checkbox/checkbox.stories.ts
```diff
@@ -0,0 +1,289 @@
+import '../../solid-components';
```
abudd1094, line 1: Could be nice to keep sizes on y axis and add descriptions for all stories to be consistent.  
Just my 2 cents not a requirement.

abudd1094 (approved):
Left small thoughts in the review.  Overall, component works, meets requirements I can see, and the stories flow pretty nicely.  Nice work!

azraefendic (approved):

EllaELSA:
one question regarding the Checkbox group horizontal. Is the distance between the single check-items also part of the comp. Because this distance is in lg and sm zu small. Should be 24px in lg and 16px in sm.

yoezlem:
> Details



> one question regarding the Checkbox group horizontal. Is the distance between the single check-items also part of the comp. Because this distance is in lg and sm zu small. Should be 24px in lg and 16px in sm.

@EllaELSA the margins between the checkboxes are inside the chechbox-group component. so to answer your question: yes, it is a part of the component.

Wanted to check it also and for me it seems right. If its horizontal and small than it has 16px:
image

and if its horizontal and large:
image

mariohamann (approved):

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,176 @@
...
+    SolidElement.styles,
+    css`
+      :host {
+        display: block;
+      }
+
+      .form-control-input--vertical ::slotted(sd-checkbox) {
+        margin-bottom: 8px;
+        display: flex;
+      }
```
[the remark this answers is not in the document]
mariohamann: But I would be fine if this is going to be done in a seperate ticket, if it take stoo much time.

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,176 @@
...
+            lg: 'text-base'
+          }[this.size]
+        )}
+        role="group"
+        aria-labelledby="label"
+      >
+        <label
+          part="form-control-label"
+          id="label"
+          class="form-control__label mb-2 hidden p-0 font-bold leading-normal text-black"
```
mariohamann, line 105: Please remove BEM classes here and use the ID as selector instead.

mariohamann (changes requested):
Sorry. Undoing approval. :D

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      .form-control-input--horizontal ::slotted(sd-checkbox) {
+        margin-right: 24px;
+      }
+
+      :host([size='sm']) .form-control-input--horizontal ::slotted(sd-checkbox) {
+        margin-right: 16px;
+      }
+
+      .form-control-input--horizontal ::slotted(sd-checkbox:last-of-type) {
```
mariohamann, line 155: Could you just check on the host if positioning is set as to omit the class `.form-control-input--horizontal`  here?

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,176 @@
...
+        margin-right: 0;
+      }
+
+      /* Label */
+      .form-control--has-label .form-control__label {
+        display: flex;
+      }
```
mariohamann, line 162: I'm pretty sure tht this will work with CS:

mariohamann (changes requested):

packages/components/src/components/checkbox-group/checkbox-group.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      :host([required]) .form-control--has-label .form-control__label::after {
+        content: '*';
+        margin-left: 2px;
+      }
```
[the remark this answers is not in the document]
DanielHargesheimer: Do you mean: should I move it from css to the template? It's the same implementation as in the Radio Group.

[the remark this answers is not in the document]
DanielHargesheimer: And as there is no required state in the Checkbox Group, I can remove it here completely. It's only relevant for the Checkbox.

yoezlem:
Wverything looks good now and can finally be merged! 🙂👍🏻

mariohamann (approved):

mariohamann:
Let's ship it!

Issue #155, opened 2023-06-13
research(dev): public NPM packages


karlbaumhauer:
Can we publish packages in NPM in a public version? Is there any regulation from UI-side which makes it more complicated?

PR #515, opened 2023-11-02
feat: ✨ add sd-input

### LINKED ISSUE #255: feat: ✨ add sd-input
## Description
As a user of the Solid Design System, I would like to use an Input component that allows me to enter text. This Input component should have the flexibility to be displayed in various ways, depending on the specified type attribute. Additionally, I should have the option to hide the default label, display a hint if needed, and have mandatory fields clearly marked with an * added to their label.

## API
https://www.figma.com/file/xSIeTnyfW2T21Uw5JgdZOg/Input?node-id=0%3A1&mode=dev

## Handoff
https://www.figma.com/file/xSIeTnyfW2T21Uw5JgdZOg/Input?node-id=6%3A4&mode=dev

## Dev notes

- [x] We start with https://github.com/solid-design-system/solid/tree/feat/add-form-elements – this already moved the needed form files into the components folder.
- [x] We take https://shoelace.style/components/input as a base, therefore move `src/_components/input` into `src/components`.
- [x] Change every style to TailwindCSS that it fits the handoff of the component. See this commit as an example: https://github.com/solid-design-system/solid/commit/a4047ee4850da0098e106765e6b8c5b05e9d28c5
- [x] API changes
  - [x] Rename `hint` to `helper-text`
  - [x] Do not change `clearable`, we will rename the boolean in design
  - [x] Do not change `disabled`, we will rename it in design

Here is a [WIP branch](https://www.figma.com/file/UPenHXItO4QfghNFBgiHCg/sd-input-V.1.0.0?type=design&node-id=29-78&mode=design) from Design with all adoptions resulting from the implementation.
#### Props

the following props only relate to design relevant props. All other props that are already provided (e.g. type, pattern, min max etc.) should not be removed.

| <ul><li>[x] </li></ul>  | `clearable` | `boolean` | `false` | shows the clear icon on the right end which clears the input field when clicked |
| <ul><li>[x] </li></ul>  | `disabled` | `boolean` | `false` | disables the input field |
| <ul><li>[x] </li></ul>  | `helper-text` | `string` | `` | shows the hint message underneath the input field |
| <ul><li>[x] </li></ul>  | `placeholder` | `string` | `` | adds plain text to the placeholder inside the input field |
| <ul><li>[ ] </li></ul>  | `error-text` | `string` | `` | adds plain text to the errormessage underneath the helper text |
| <ul><li>[x] </li></ul>  | `label` | `string` | `` | adds plain text to the label above the input field |
| <ul><li>[x] </li></ul>  | `size` | `lg` / `md` / `sm` | `lg` | defines the height of the input field |

#### Parts
- [x]  all existing one's from Shoelace
- [ ] `error-text`

#### Slots
  - [x] `label` - slot for the input field label
  - [x] `helper-text` - slot for the helper text
  - [ ] `error-text` - slot for the error message
  - [ ] `left` - slot to add prefix content to the start of the input field
  - [ ] `right` - slot to add suffix content to the end of the input field
  - [x] `placeholder` - slot for the placeholder of the input field
  - [ ] Sizes & Icons - sizes, icons, clearable 
  - [ ] Descriptions - helper text, error text, placeholder & label
#### Samples
  - [ ] `tooltip` - An input field with a sd-interactive (using info-circle) added to the label
  - [ ] [stepper](https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/branch/UPenHXItO4QfghNFBgiHCg/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=1211-14461&mode=design&t=APz9feeqDB0nyiHr-11) 

- [x] @mariohamann / @coraliefeil if we activate the tooltip, how does the content for the tooltip will be added? dont we need a slot for this?
- [ ] @mariohamann / @coraliefeil is the `icon-right` property really `true` by default? => no/maybe. @van-nguyen-ht clarifies that and adds a comment here if needed: the field input with the icon-right true by default is the auto-suggest with the search icon.
- [ ] @coraliefeil the `message` property is not available in the playground. Do you know why? => @van-nguyen-ht verifies it. ??
- [x] @solid-design-system/design what is the cursor property about? This seems like a pure design property (might just need an asterisk). => yes. @van-nguyen-ht is changing it.
- [x] @coraliefeil / @van-nguyen-ht pls make sure the figma file is updated or we have a link to the respective branch in this ticket. There are several mismatching information in this ticket and the current figma file regarding the API.
- [x] @mariohamann pls go through my changes and check if I missed something.
- [x] @mariohamann pls define the samples we need for this component.
- [x] @mariohamann pls define which parts from shoelace we need and change the list above accordingly.

## DoR (Definition of Ready)
- [x] This user story provides value to the product.
- [x] The development team has estimated this user story.
- [ ] The user story is clear and well-defined.
- [x] Any dependencies related to this user story have been identified.

## DoD (Definition of Done)
- [ ] Documentation for this Input component has been created or updated if applicable.
- [ ] A migration guide is available or updated if applicable.
- [ ] Relevant end-to-end tests for features, accessibility, and bug fixes are created or updated.
- [ ] Relevant stories for features and accessibility are created or updated.
- [ ] Implementation successfully works on the `feature` branch.

abudd1094:
## Description:
Added sd-input component and samples. Closes #255

Please read the TODO comments.  They are meant to call attention to certain lines where there is not full clarity.

## Definition of Reviewable:
- [x] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [x] Implementation works successfully on feature branch


abudd1094:
The following test adapted from Shoelace succeeds locally and fails in the pipeline:
`  describe('when using the getFormControls() function', () => {
    it('should return both native and Solid form controls in the correct DOM order', async () => {
      const el = await fixture<HTMLFormElement>(html`
        <div>
          <input type="text" name="a" value="1" form="f1" />
          <sd-input type="text" name="b" value="2" form="f1">2</sd-input>
          <form id="f1">
            <input type="hidden" name="c" value="3" />
            <input type="text" name="d" value="4" />
            <sd-input name="e" value="5">5</sd-input>
            <textarea name="f">6</textarea>
            <textarea name="g">7</textarea>
            <input type="checkbox" name="h" value="8" />
          </form>
          <input type="text" name="i" value="9" form="f1" />
          <sd-input type="text" name="j" value="10" form="f1"></sd-input>
        </div>
      `);
      const form = el.querySelector<HTMLFormElement>('form')!;
      const formControls = getFormControls(form); // eslint-disable-line

      expect(formControls.length).to.equal(10); // eslint-disable-line
      expect(formControls.map((fc: HTMLInputElement) => fc.value).join('')).to.equal('12345678910'); // eslint-disable-line
    });
  });`
  
  The collection constant should find 3 sd-inputs but finds none when run in the pipeline.  There may be some deeper logic that was not adapted to Solid from Shoelace?  I can't figure it out.  
  `/**
 * Returns all form controls that are associated with the specified form. Includes both native and Solid form
 * controls. Use this function in lieu of the `HTMLFormElement.elements` property, which doesn't recognize Solid
 * form controls.
 */
export function getFormControls(form: HTMLFormElement) {
  const rootNode = form.getRootNode() as Document | ShadowRoot;
  const allNodes = [...rootNode.querySelectorAll('*')];
  const formControls = [...form.elements];
  const collection = formCollections.get(form);
  const solidFormControls = collection ? Array.from(collection) : [];

  // To return form controls in the right order, we sort by DOM index
  return [...formControls, ...solidFormControls].sort((a: Element, b: Element) => {
    if (allNodes.indexOf(a) < allNodes.indexOf(b)) return -1;
    if (allNodes.indexOf(a) > allNodes.indexOf(b)) return 1;
    return 0;
  });
}`
  
  The test is being run in the pipeline, which is different than the local environment.  There it does not find any sd-inputs... are they somehow not registered?  Is some step missing?  @mariohamann it does not make sense to spend any more time on this for me.  Please advise, I will comment this test out for now.

abudd1094:
@solid-design-system/core-development 
I propose we designate a format for samples, otherwise it will get real messy real quick!
image

I copied the styles that @Vahid1919 used for the button samples here...

coraliefeil:
Design review done. See comments on chromatic. Pls reassign when adjusted. Thanks a lot <3

abudd1094:
image
This TS error seems to not show up for @Vahid1919 or in Shoelace itself (which implemented both the controller logic and this specific use case).  I took a deep dive into it and believe it is caused by the use of the spread operator for args in the type definition for arg params... @mariohamann It does not seem to prevent the verify or pipeline from running, so I am moving on.  If you are concerned or curious, contact me.

abudd1094:
@coraliefeil first round of UI tests have been addressed.  Sounds like we have a couple decisions to make.

Regarding the types, Coralie would like to remove the following for now:
- date
- datetime
- search

Makes sense to me because we don't yet have the solid counterparts yet.  I would say we comment this stuff out because it will become relevant again soon (although that is not the 'cleanest' approach it may be most conducive to progress).

I am also still wondering about password related stuff.  I have the icons replaced with 'show' / 'hide' text, shall I just remove this completely for now (no show/hide button) ?  Probably design will not like the look (:

mariohamann:
I'm strongly against removing these types. I have the feeling, that regarding datepickers we'll go more into direction of https://flatpickr.js.org/ and providng a theme for that. So there maybe will never be a component for it.

But still an input field should just work and validate against that. It might even be, that people use it in combination with flatpickr or whatever. We shouldn't prevent developers to use that.

yoezlem:
I agree with Mario - we should definitely not start excluding types, especially when we had already discussed that the datepicker would be large enough to warrant its own package.

packages/components/package.json
```diff
@@ -133,7 +133,7 @@
     "semantic-release": "^19.0.5",
     "semantic-release-monorepo": "^7.0.5",
     "sinon": "^17.0.0",
-    "storybook": "^7.3.2",
+    "storybook": "^7.5.1",
```
abudd1094, line 136: Will this be an issue?? 😅

packages/components/src/components/button/button.ts
```diff
@@ -24,7 +24,7 @@ import type { SolidFormControl } from '../../internal/solid-element';
  * @event sd-blur - Emitted when the button loses focus.
  * @event sd-focus - Emitted when the button gains focus.
  * @event sd-invalid - Emitted when the form control has been checked for validity and its constraints aren't satisfied.
-
+
```
abudd1094, line 27: -_-

coraliefeil:
Left some comments on chromatic. Pls reassign when adjusted.

Not sure how to proceed with type that should stay in the comp needing styling -> can only open an issue and provide skinning examples ...

packages/components/src/components/input/input.ts
```diff
@@ -388,61 +403,129 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
-      'input--disabled': this.disabled,
-      'input--focused': this.hasFocus,
-      'input--empty': !this.value,
-      'input--no-spin-buttons': this.noSpinButtons,
-      'input--is-firefox': isFirefox
-    })}
+            class=${cx(
+              'px-4 rounded flex flex-row items-center rounded-[4px]',
+              // Vertical Padding
+              this.size === 'lg' ? 'py-2' : 'py-1',
```
yoezlem, line 495: Is there a specific reason why you've chosen a completely different query here compared to the other components? Doesn't it work like this:
```
{
  /* sizes */
  sm: 'py-1',
  md: 'py-1',
  lg: 'py-2'
}[this.size]
```

For me this one is more readable.

packages/components/src/components/input/input.ts
```diff
@@ -388,61 +403,129 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
+      readonly: 'border-neutral-800',
+      activeInvalid: 'border-error border-2',
+      activeValid: 'border-success border-2',
+      active: 'border-primary border-2',
+      invalid: 'border-error',
+      valid: 'border-success',
+      default: 'border-neutral-800'
+    }[inputState];
+
+    const iconMarginLeft = this.size === 'sm' ? 'ml-1' : 'ml-2';
```
yoezlem, line 463: Please write iconMarginLeft like you did in iconSize:
`const iconMarginLeft = {
      sm: 'ml-1',
      md: 'ml-2',
      lg: 'ml-2'
    }[this.size];`

yoezlem:
The clear icon has a different color in Figma compared to your code.

Storybook:
image

Figma:
image

packages/components/package.json
```diff
@@ -85,25 +85,25 @@
...
+    "@storybook/addon-interactions": "^7.5.1",
+    "@storybook/addon-links": "^7.5.1",
+    "@storybook/addon-mdx-gfm": "^7.5.1",
+    "@storybook/addons": "^7.5.1",
+    "@storybook/blocks": "^7.5.1",
+    "@storybook/jest": "^0.2.3",
+    "@storybook/testing-library": "^0.2.2",
+    "@storybook/theming": "^7.5.1",
+    "@storybook/web-components": "^7.5.1",
+    "@storybook/web-components-vite": "^7.5.1",
```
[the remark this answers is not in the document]
abudd1094: https://github.com/solid-design-system/solid/issues/255#issuecomment-1781214142

As I commented, Storybook has a bug that is fixed in this version.  I have asked what is the best way to do this.  Shall we update Storybook in a separate PR?  Without it, we will have this error present.

packages/components/src/components/navigation-item/navigation-item.stories.ts
```diff
@@ -317,7 +317,7 @@ export const Slots = {
 };
 
 /**
- * Use the 'base', 'content', 'chevron', and 'description' part selectors to customize the navigation item.
+ * Use the `base`, `content`, `chevron`, and `description` part selectors to customize the navigation item.
```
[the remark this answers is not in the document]
abudd1094: Shall I make a separate PR?

packages/components/src/components/input/input.ts
abudd1094: One line vs. 5 lines, 2 conditions vs. 3 conditions.  If you prefer class maps, I will use this as a guideline.

packages/components/package.json
abudd1094: @yoezlem

abudd1094:
> The clear icon has a different color in Figma compared to your code.
> 
> Storybook: image
> 
> Figma: image


> The clear icon has a different color in Figma compared to your code.
> 
> Storybook: image
> 
> Figma: image

Please read my TODO comments. My questions are being overlooked and I need your feedback to resolve these points.

input.ts line 569:
`<slot name="clear-icon">
  
  <sd-icon
    class="text-neutral-400"
    library="global-resources"
    name="system/closing-round"
  ></sd-icon>
</slot>`

There is currently a misalignment on icon fills:
image

@coraliefeil is correct that the fill property is used for SVG elements, however no other components are doing this. 'icon-fill-neutral-400' when used as a class on sd-icon has no effect.  The [CSS fill property](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill) must be applied directly to the `path`element and this should be implemented in `sd-icon`if we wish to use the 'icon-fill' tokens provided by design.  At the moment, this does not work.

In every other component involving icons (e.g. sd-button) the 'text' color tokens are used to style icons.  I followed this example.  In the snippet above, if I apply the token based class 'text-neutral-500', the icon will be styled accordingly.  However, 'text-neutral-400' is not recognized as a class and therefore applies no styles.  As mentioned above, 'icon-fill-neutral-400' will also have no effect because it is not correctly applied to the path elements in 'sd-icon'.

@mariohamann it is not 100% clear to me how you are converting tokens to classes in create-theme.cjs, it is minimally documented.  I can inspect further or better yet you can give me a short introduction or resolve the missing class directly.

At this point, I have raised the awareness, please support me with a decision @solid-design-system/core-development

abudd1094:
@mariohamann I would like your feedback directly before I implement the following:
image

I also insist that we update the requirements above (in the first comment) as we resolve these questions so I can have full clarity on the final expectations.

coraliefeil:
@abudd1094 
[Eye Icon](https://www.figma.com/file/KNIe24a9qGUmpZ4ntnPN5P/Solid-DS-%E2%80%93-Icon-Asset-Library?type=design&node-id=1335-2305&mode=design&t=m2UbAl3VFkMRvKyI-11)
[Eye Crossed Out Icon](https://www.figma.com/file/KNIe24a9qGUmpZ4ntnPN5P/Solid-DS-%E2%80%93-Icon-Asset-Library?type=design&node-id=1335-2305&mode=design&t=m2UbAl3VFkMRvKyI-11)

abudd1094:
> @mariohamann I would like your feedback directly before I implement the following: image
> 
> Should we update the requirements above (in the first comment) as we resolve these questions so I can have full clarity on the final expectations?

@solid-design-system/core-development 
image
It is not possible to hide the datepicker icon in firefox.  Therefore this request is further complicated.
A common workaround is to use the 'text' type for dates, which is also what flatpickr does.  Then we lose the input formatting and default validation for date types.  Tradeoffs.  I am blocked without a decision.  What do you propose?

abudd1094:
@solid-design-system/core-development @solid-design-system/design 
image
Some system icons include the version 1.1, I don't know where this comes from.  Is it the library version?  For the newly added system icons shall I include version="1.2.1" (the next anticipated release)?  Please advise or I will exclude a version.

mariohamann:
> @solid-design-system/core-development @solid-design-system/design image Some system icons include the version 1.1, I don't know where this comes from. Is it the library version? For the newly added system icons shall I include version="1.2.1" (the next anticipated release)? Please advise or I will exclude a version.

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version :) This is a SVG thing, you can remove it.

mariohamann:
> > @mariohamann I would like your feedback directly before I implement the following: image
> > Should we update the requirements above (in the first comment) as we resolve these questions so I can have full clarity on the final expectations?
> 
> @solid-design-system/core-development image It is not possible to hide the datepicker icon in firefox. Ignore the eye icon. Therefore this request is further complicated. A common workaround is to use the 'text' type for dates, which is also what flatpickr does. Then we lose the input formatting and default validation for date types. Tradeoffs. I am blocked without a decision. What do you propose?

Wow. What do you think about hiding our icon for non FireFox-Browsers? Is not compliant but seems like the easiest compromise.

packages/components/src/components/navigation-item/navigation-item.stories.ts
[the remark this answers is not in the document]
yoezlem: No just keep in mind, to separate it the next time.

packages/components/package.json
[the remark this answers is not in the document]
yoezlem: No just keep in mind, to separate it the next time.

packages/components/package.json
yoezlem: What do you mean?
If you mean to separate it, you don't have to do it in this PR. Just keep in mind to do it next time.

mariohamann:
So my thoughts on that:

## Dev perspective
1. type=date does allows min/max, validation, formatting and shows native datepickers, what is especially nice for mobile users.
2. If we go the "text" route, there's not a inputmode that reflects the native date picker (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode).
3. A full fledged, pre-designed datepicker like when using flatpicker needs implementation effort and raises bundle sizes a LOT.

This should be a decision teams using our design system should have a choice for. TBH: In 99% a standard date picker will be just enough.

For the rest, using flatpicker or similar would be fine too implement, especially if it's a critical part of the implementation.

## Design perspective
1. The UI calendar icon should show up instead of the native one.
2. The branded date-picker should be used instead of the native one.

## Suggestion
I'm fine how we will handle error handling and will implement errors inline, instead of using the browser's behaviour, because this won't cost us a lot in terms of code and effort. Removing "date=type" just because of FireFox showing its own calendar icon is not proportionate to the losses we would have in development and forcing to a maybe way to oversized implementation for most cases.

To get a good compromise I would suggest:

1. `date=type` shows the UI calendar icon, but hides it on FireFox. This leads to always showing one calendar icon and apparently is the only way to overcome the problem of showing two icons. Yes, the calendar icon will at another position, but we can't override the browser here.
2. We show a sample later which shows using flatpicker in combination with `date=text` and manually setting the calendar icon in the slot. So teams using flatpicker will be able to show only the calendar icon, no native ones, as flatpickr will (hopefully) handle validation etc.

yoezlem:
@abudd1094 should the stepper (sample) not do something?
Just tried do push the plus button, but nothing happens.
image

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,642 @@
...
+        'minlength',
+        'maxlength',
+        'min',
+        'max',
+        'form',
+        'name',
+        'required',
+        'pattern',
+        'step'
+      ]
```
mariohamann, line 605: In think you could whitelist via include. :)

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,642 @@
...
+              id="stepDownButton"
+              @click=${() => step('down')}
+              class="ml-4 scale-[1.714] inline-flex items-center text-neutral-500"
+            >
+              <sd-icon library="global-resources" name="system/minus-round"></sd-icon>
+            </button>
+            <button id="stepUpButton" @click=${() => step('up')} class="ml-4 scale-[1.714] inline-flex items-center">
+              <sd-icon library="global-resources" name="system/plus-round"></sd-icon></button
+          ></span>
+        </sd-input>
```
mariohamann, line 637:
```suggestion
<sd-input id="numericInput" type="number" min="0" class="w-[231px]">
  <span slot="right" class="text-sm inline-flex items-center gap-4">
    <span class="text-neutral-700">EUR</span>
    <button
      onclick="document.getElementById('numericInput').stepDown()"
      class="scale-[1.714] inline-flex items-center text-neutral-500"
    >
      <sd-icon library="global-resources" name="system/minus-round"></sd-icon>
    </button>
    <button
      onclick="document.getElementById('numericInput').stepUp()"
      class="scale-[1.714] inline-flex items-center"
    >
      <sd-icon library="global-resources" name="system/plus-round"></sd-icon>
    </button>
  </span>
</sd-input>
```

Could you try something like this?

packages/components/src/components/input/input.ts
```diff
@@ -388,61 +403,129 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
+      right: this.hasSlotController.test('right')
+    };
+
+    // States
+    const hasLabel = this.label ? true : !!slots['label'];
+    const hasHelpText = this.helpText ? true : !!slots['helpText'];
+    const hasClearIcon = this.clearable && !this.readonly && (typeof this.value === 'number' || this.value.length > 0);
+    const hasValidationAttr = this.required || !!this.pattern || !!this.minlength || !!this.maxlength;
+    const isInvalid = hasValidationAttr && !this.checkValidity();
+    const isValid = hasValidationAttr && this.checkValidity();
```
saemik94, line 421: I have just noticed that the basically same constant appears twice here. Moreover, if `hasValidationAttr = false`, then `isInvalid = false` and `isValid = false` too, which makes no sense.

packages/components/src/components/input/input.stories.ts
abudd1094: didn't realize that was an option

packages/components/src/components/input/input.stories.ts
abudd1094: That is what I originally did, it will not apply conditional styles or currency format which reduces the quality of the story.  

@yoezlem clicking the step buttons fires my step function which I took the time to implement for the sake of this sample and furthermore documented via comments:
https://github.com/solid-design-system/solid/assets/39494579/34b6d104-11e1-4bd6-bc06-23a622ae33b9

Is it not working for you?  If not I would need details to further understand.  We can either refine that logic or remove it and have a simple story with no formatting or conditional styles.  Either is fine for me.

packages/components/src/components/input/input.ts
abudd1094: isInvalid is used to hide or show the "risk" icon
isValid is used to hide or show the "success" icon
image

IF there is no validation attribute, then there shouldn't be either type of validation icon present.  This makes sense to me.  How would you change this to make it more readable / understandable? I'm open to suggestions.  However, the 'success' state does not really make sense to me.  I included a TODO comment which was intended to stimulate conversation:

line 630:


I'm a little frustrated that my comments are still being ignored and I am only offered cherry-picked feedback here.  I will bring this up in the daily tomorrow.

packages/components/src/components/input/input.stories.ts
mariohamann: Ah, just saw now the method. It's super nice.

if you switch to docs where all stories are it's not working.

Does it work with onclick instead of @click? Or do we maybe have elements with the same id or similar?

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,675 @@
...
+  };
+
+  // function to adjust classlist of stepDownButton
+  const adjustClasslist = () => {
+    if (inputEl.value === '0') {
+      addClass(stepDownButton, 'text-neutral-500');
+    } else {
+      removeClass(stepDownButton, 'text-neutral-500');
+    }
+  };
```
mariohamann, line 593: The problem for the docs might be, that this is maybe not available inside the story.

Could you insert all this JS inside the story? This is especially important as people should be able to see, what you're doing and if you have it outside the story that doesn't work. :)

packages/components/src/components/input/input.stories.ts
mariohamann: I just realized that the problem for the docs might be, that your function is maybe not available inside the story.


CleanShot 2023-11-09 at 08 24 25@2x

That's why I wasn't really aware of what the method is doing. 
 So: Could you insert all this JS inside the story? This is especially important as people should be able to see, what you're doing and if you have it outside the story that doesn't work. :)

packages/components/src/components/input/input.stories.ts
mariohamann, line 593: Could you use `sd-interactive` here on the button? https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/styles-sd-interactive--docs

Instead of toggling the classes you then have just to toggle the disabled attribute which would especially help in regard of accessibility. :)

mariohamann: BTW: I really like the story/sample.

packages/components/src/components/input/input.stories.ts
yoezlem: Exactly! It works only within the documentation sample, not in the actual story. Strange behavior 🤔

yoezlem:
> So my thoughts on that:
> 
> ## Dev perspective
> 1. type=date does allows min/max, validation, formatting and shows native datepickers, what is especially nice for mobile users.
> 2. If we go the "text" route, there's not a inputmode that reflects the native date picker (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) and we lose eveything else mentioned above, too.
> 3. A full fledged, pre-designed datepicker like when using flatpicker needs implementation effort and raises bundle sizes a LOT.
> 
> IMO: This should be a decision teams using our design system should have a choice for. TBH: In 99% a standard date picker will be just enough.
> 
> For the rest, using flatpicker or similar would be fine too implement, especially if it's a critical part of the implementation.
> 
> ## Design perspective
> 1. The UI calendar icon should show up instead of the native one.
> 2. The branded date-picker should be used instead of the native one.
> 
> ## Problem
> I'm fine how we will handle error handling and will implement errors inline, instead of using the browser's behaviour, because this won't cost us a lot in terms of code and effort. Removing "date=type" just because of FireFox showing its own calendar icon is not proportionate to the losses we would have in development and forcing to a maybe way to oversized implementation for most cases.
> 
> ## Suggestion
> To get a good compromise I would suggest:
> 
> 1. `date=type` shows the UI calendar icon, but hides it on FireFox. This leads to always showing one calendar icon and apparently is the only way to overcome the problem of showing two icons. Yes, the calendar icon will at another position, but we can't override the browser here.
> 2. We show a sample later which shows using flatpicker in combination with `date=text` and manually setting the calendar icon in the slot. So teams using flatpicker will be able to show only the calendar icon, no native ones, as flatpickr will (hopefully) handle validation etc.

I like Mario's suggestion, let's go ahead with that 👍🏻

packages/components/src/components/input/input.ts
```diff
@@ -36,8 +36,8 @@ const isFirefox = isChromium ? false : navigator.userAgent.includes('Firefox');
  * @dependency sd-icon
  *
  * @slot label - The input's label. Alternatively, you can use the `label` attribute.
- * @slot prefix - Used to prepend a presentational icon or similar element to the input.
- * @slot suffix - Used to append a presentational icon or similar element to the input.
+ * @slot left - Used to prepend a presentational icon or similar element to the input.
+ * @slot right - Used to append a presentational icon or similar element to the input.
```
yoezlem, line 40: I think the naming if you're only on Storybook, is confusing. Would suggest to be clearer if we also add an icon.
So, if this slot adds only an icon on the left or right side.. could we rename it to left-icon/right-icon or prefix-icon/suffix-icon?

packages/components/src/components/input/input.stories.ts
abudd1094: This is what I originally did, however then step is not recognized.
image

I tried several workarounds and published the result because it was working in my story.  I'm not sure how much more time that is worth.  It will be much easier to implement in an actual counter component.  

I'll try a couple more things.

packages/components/src/components/input/input.ts
mariohamann: It's not meant only for icons. You can add e. g. `EUR` or `mm` or something, too.

I would have liked prefix and suffix more, but as we already (unfortunately) have `icon-left` and `icon-right` this is the consequence. ^^

mariohamann: As this a slot name, it is ok.

yoezlem: Can we then modify the text a bit and write `icon, EUR, or a similar element`? I can imagine that others might be confused as well, thinking that only an icon fits there. Without your example with EUR and mm, I wouldn't know what a similar element could be.

packages/components/src/components/input/input.ts
yoezlem: @saemik94

packages/components/src/components/input/input.ts
mariohamann: Then maybe just `text, icons or arbitrary content` to the input?

coraliefeil:
> > > The clear icon has a different color in Figma compared to your code.
> > > Storybook: image
> > > Figma: image
> > 
> > 
> > > The clear icon has a different color in Figma compared to your code.
> > > Storybook: image
> > > Figma: image
> > 
> > 
> > Please read my TODO comments. My questions are being overlooked and I need your feedback to resolve these points.
> > input.ts line 569: `
> > <sd-icon class="text-neutral-400" library="global-resources" name="system/closing-round"
> > > 
> > 
> > 
> > `
> > There is currently a misalignment on icon fills: image
> > @coraliefeil is correct that the fill property is used for SVG elements, however no other components are doing this. 'icon-fill-neutral-400' when used as a class on sd-icon has no effect. The [CSS fill property](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill) must be applied directly to the `path`element and this should be implemented in `sd-icon`if we wish to use the 'icon-fill' tokens provided by design. At the moment, this does not work.
> > In every other component involving icons (e.g. sd-button) the 'text' color tokens are used to style icons. I followed this example. In the snippet above, if I apply the token based class 'text-neutral-500', the icon will be styled accordingly. However, 'text-neutral-400' is not recognized as a class and therefore applies no styles. As mentioned above, 'icon-fill-neutral-400' will also have no effect because it is not correctly applied to the `path` elements in `sd-icon`.
> > @mariohamann it is not 100% clear to me how you are converting tokens to classes in create-theme.cjs. I can inspect further and try to resolve it myself or I can use your assistance 😄
> 
> I can confirm Alec's statement. I just tried it myself and encounter the same issues. "text-neutral-400" is not suggested as a token, but 500 is suggested. @mariohamann

Ok, if - for now - you can only use text color tokens ... then feel free to use the available one text-neutral-500 ... not a big deal in that case. Thanks

coraliefeil:
Pls reassign when I should review changes

packages/components/src/components/input/input.ts
saemik94: Ahh, now I understand what's going on, it's more difficult than I thought. Thanks for the explanation

abudd1094:
Ok, more updates are pushed.

1. help-text is now text-neutral-700 for all states and variants.  This is not documented in Figma, refer to comment from Coralie on Chromatic:
image

2. As per the discussion, text-neutral-400 is not available.  text-neutral-500 is used instead, TODO comment left:
image

3. Do we want to leave the 'message' slot in the component or remove it?
image

4. Solid icon styling.  For the calendar and clock, this is an open question.  Is the order correct and how would you like to apply colors in the various states to these icons @coraliefeil?  Currently nothing is implemented so it inherits color.
image

5. Regarding differences in the input types, I have followed Mario's and Coralie's suggestions.  Solid clock icon is used for the time type.  Solid calendar icon is used for the date and datetime-local types, except for firefox, which keeps the default browser datepicker instead (impossible to hide without ugly hacks).  flatpickr will not allow someone to use sd-input with type 'date' or 'datetime-local' it will foce the input to 'text' type and thus voids the built in browser formatting and validation.  This is an open topic for us and how we would like to handle form validation and dates in our library.

I await new feedback.
@solid-design-system/current-sds-team

coraliefeil:
Pls reassign when last comments have been implemented. THX

coraliefeil:
Approved. Thank you for pushing through! Really a tough component. Appreciate all your helpful questions. <3

karlbaumhauer:
As there have been so many reviews already, I will remove myself from the assignees. I dont feel, I could contribute anything meaningful to the PR and especially dont want to block the merge anymore. 

If you need anything specific from me about this PR, pls reach out to me directly.

packages/components/src/components/input/input.ts
yoezlem: @abudd1094  Can you change the text for left & right as suggested by Mario?

yoezlem:
> Ok, more updates are pushed.
> 
> 1. help-text is now text-neutral-700 for all states and variants.  This is not documented in Figma, refer to comment from Coralie on Chromatic:
>    image
> 2. As per the discussion, text-neutral-400 is not available.  text-neutral-500 is used instead, TODO comment left:
>    image
> 3. Do we want to leave the 'message' slot in the component or remove it?
>    image
> 4. Solid icon styling.  For the calendar and clock, this is an open question.  Is the order correct and how would you like to apply colors in the various states to these icons @coraliefeil?  Currently nothing is implemented so it inherits color.
>    image
> 5. Regarding differences in the input types, I have followed Mario's and Coralie's suggestions.  Solid clock icon is used for the time type.  Solid calendar icon is used for the date and datetime-local types, except for firefox, which keeps the default browser datepicker instead (impossible to hide without ugly hacks).  flatpickr will not allow someone to use sd-input with type 'date' or 'datetime-local' it will foce the input to 'text' type and thus voids the built in browser formatting and validation.  This is an open topic for us and how we would like to handle form validation and dates in our library.
> 6. As far as a 'success' state for validation, I'm not sure if my implementation is what is expected.  Please have a look.  @saemik94 made a valid point above that he did not understand the conditions for 'hasValidationAttr', 'isInvalid', and 'isValid' which I used in the render method.  Is it also confusing for you?  How would you recommend changing it if so?
> 7. I updated the stories to resemble Shoelace, they have a pretty readable presentation.  The sample is also fixed.  I think it can be refined a bit further.  Maybe we should replace the Validation story with a types story?  Any thoughts?
> 
> I await new feedback. @solid-design-system/current-sds-team

1. @coraliefeil do you need to document this to Figma too?
2. Thx for the comment, do we need to create also a follow up ticket for that? @mariohamann 
3. I would leave it as it is, but we can also create a follow-up ticket (as I mentioned in Point 2), remove it here for now, and specify in the follow-up ticket that it needs to be added back. However, I find it unnecessary, so I suggest we leave it in.
4. For me it looks how it should!
5.  I'm not sure if we should postpone this discussion until we have a "proper" datepicker. Documentation will likely be crucial there so that the user understands what they can do with <input type="date"> and Flatpickr <input type="text">.
6. I find it understandable, and I'm not sure how it could be explained in more detail.
7. can't we do both?

coraliefeil:
1. All adjustments while coding are documented in the [figma branch](https://www.figma.com/file/UPenHXItO4QfghNFBgiHCg/sd-input-V.1.0.0?type=design&node-id=29-78&mode=design)
2. I opened up a ticket tackling color tokens icon-fill / text - foreground [#155](https://github.com/solid-design-system/figma/issues/155)
3. – not design related
4. Fine ... 
5. We’ll talk next week. Postponing sounds good.
6. We could add an inline notification or just a text on top telling the user to fill out all required fields to see valid states
7. –

packages/components/src/components/input/input.ts
```diff
@@ -89,39 +96,40 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
+  /** The input's size. */
+  @property({ reflect: true }) size: 'lg' | 'md' | 'sm' = 'lg';
+
+  /**
+   * Tells the browser what type of data will be entered by the user, allowing it to display the appropriate virtual
+   * keyboard on supportive devices.
+   */
+  @property() inputmode: 'none' | 'text' | 'decimal' | 'numeric' | 'tel' | 'search' | 'email' | 'url';
 
+  // --------------------------- GENERIC --------------------------- //
```
mariohamann, line 108: Do you mean by "Generic" that these are things which are inherited by the original input? :) Not sure if I like it, but don't have strong feelings on it. If you leave it, maybe we can have a better wording like "original input attributes" or similar.

packages/components/src/components/input/input.ts
```diff
@@ -471,68 +549,190 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
-        ? html`
+            >
+            <!-- TODO: substitute text-neutral-400 for text-neutral-500 when available -->
+            ${
+              hasClearIcon
+                ? html`
                     <button
                       part="clear-button"
-                      class="input__clear"
+                      class=${cx('input__clear flex justify-center ', iconMarginLeft)}
```
mariohamann, line 559:
```suggestion
                      class=${cx('flex justify-center ', iconMarginLeft)}
```

packages/components/src/components/input/input.ts
```diff
@@ -388,61 +403,124 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
+                    class=${cx('input__left inline-flex', this.size === 'sm' ? 'mr-1' : 'mr-2', iconColor, iconSize)}
+                  ></slot>`
+                : ''
+            }
             <input
               part="input"
               id="input"
-              class="input__control"
+              class=${cx(
+                'input__control min-w-0 flex-grow focus:outline-none bg-transparent placeholder-neutral-700',
```
mariohamann, line 520:
```suggestion
                'min-w-0 flex-grow focus:outline-none bg-transparent placeholder-neutral-700',
```

packages/components/src/components/input/input.ts
```diff
@@ -388,61 +403,124 @@ export default class SdInput extends SolidElement implements SolidFormControl {
...
+              isFirefox && 'input--is-firefox'
+            )}
           >
-            <slot name="prefix" part="prefix" class="input__prefix"></slot>
+            ${
+              slots['left']
+                ? html`<slot
+                    name="left"
+                    part="left"
+                    class=${cx('input__left inline-flex', this.size === 'sm' ? 'mr-1' : 'mr-2', iconColor, iconSize)}
```
mariohamann, line 512:
```suggestion
                    class=${cx('inline-flex', this.size === 'sm' ? 'mr-1' : 'mr-2', iconColor, iconSize)}
```

packages/components/src/components/input/input.ts
```diff
@@ -54,24 +54,31 @@ const isFirefox = isChromium ? false : navigator.userAgent.includes('Firefox');
...
+    'help-text',
+    'label',
+    'left',
+    'right',
+    'message',
+    'placeholder'
+  );
   private readonly localize = new LocalizeController(this);
 
   @query('.input__control') input: HTMLInputElement;
```
mariohamann, line 79:
```suggestion
  @query('#input') input: HTMLInputElement;
```

packages/components/src/components/input/input.ts
```diff
@@ -456,7 +534,7 @@ export default class SdInput extends SolidElement implements SolidFormControl {
               max=${ifDefined(this.max)}
               step=${ifDefined(this.step as number)}
               .value=${live(this.value)}
-              autocapitalize=${ifDefined(this.type === 'password' ? 'off' : this.autocapitalize)}
+              autocapitalize=${ifDefined(this.type === 'password' ? 'off' : this.autocapitalize)}
```
mariohamann, line 537:
```suggestion
              autocapitalize=${ifDefined(this.type === 'password' ? 'off' : this.autocapitalize)}
```

packages/components/src/components/input/input.ts
```diff
@@ -471,68 +549,190 @@ export default class SdInput extends SolidElement implements SolidFormControl {
               @keydown=${this.handleKeyDown}
               @focus=${this.handleFocus}
               @blur=${this.handleBlur}
-            />
-
-            ${hasClearIcon
-        ? html`
+            >
+            <!-- TODO: substitute text-neutral-400 for text-neutral-500 when available -->
```
mariohamann, line 553: Did you open up a ticket for that? That would be great.

packages/components/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,676 @@
...
+          margin-bottom: 16px;
+          padding: 16px;
+          background: #e0e0e0;
+          text-align: left;
+          font-size: 14px;
+          font-weight: bold;
+          width: 100%;
+          box-sizing: border-box;
+        }
+      </style>
```
mariohamann, line 630: Could you move this into a decorator, so that we don't see it in the story's code? :)

packages/components/src/components/input/input.ts
mariohamann: https://github.com/solid-design-system/figma/issues/155

Just read that there already is an issue. :)

mariohamann:
Okay, from what I read everything from a concept perspective is resolved.

Props for the stories. This is really extensive and helpful.

mariohamann:
Oh, one last thing before I forget it: Please add transitions to the colors especially on hover, but maybe even focus and error. We try to avoid hard color changes.

packages/components/src/components/input/input.ts
abudd1094: it was a label for me to separate the props categorically while developing, will remove it!

abudd1094:
> Oh, one last thing before I forget it: Please add transitions to the colors especially on hover, but maybe even focus and error. We try to avoid hard color changes.

I've added transition-all to the base and border parts, adding it to input yields some strange behaviors (flickering of border around native input).  Give a shout if it is still lacking

mariohamann (approved):
Really appreciate your effort here. 🙏🏻

PR #575, opened 2023-11-16
feat: ✨ update sd-accordion meeting WCAG criteria / design refresh

### LINKED ISSUE #407: feat: ✨ update sd-accordion meeting WCAG criteria / design refresh
## User Story
As a developer of the Solid Design System, I would like to meet the WCAG criteria in the sd-accordion, as well as used the latest version from the Figma component, so that development is aligned with design.

### Suggested Solution
Update the component regarding the latest designs from https://github.com/solid-design-system/solid/issues/558

Please have in mind, that the green line of an opened accordion grows when hovering.

Image

Check out the navigation item regarding transition. (It should include one.)

## DoR
- [x] Item has business value

## DoD

Vahid1919:
## Description:
sd-accordion refresh. Closes #407.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/accordion/accordion.ts
+++ b/packages/components/src/components/accordion/accordion.ts
@@ -150,13 +150,12 @@ export default class SdAccordion extends SolidElement {
 
   render() {
     return html`
-      <div part="base" class="outline outline-1 outline-neutral-400 -outline-offset-1">
+      <div part="base" class="border-y-[1px] border-neutral-400">
         <header
           part="header"
           id="header"
           class=${cx(
-            'flex text-base gap-4 font-bold items-center cursor-pointer select-none px-4 py-3 focus-visible:focus-outline',
-            this.open ? 'bg-white text-accent hover:bg-neutral-200' : 'text-primary bg-neutral-100 hover:bg-neutral-200'
+            'flex text-base gap-4 font-bold items-center cursor-pointer select-none px-4 py-3 focus-visible:focus-outline text-primary hover:bg-neutral-200 relative group'
           )}
           role="button"
           aria-expanded=${this.open ? 'true' : 'false'}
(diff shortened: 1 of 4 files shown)


karlbaumhauer (approved):

abudd1094 (approved):

coraliefeil:
Pls reassign when updated. Thanks!

Vahid1919:
@coraliefeil Thanks for the review :) 

For the background story. You're right, It was implemented like a variant. 

Since it is supposed to be a guideline only, I am now wondering if it should be a story at all and if the Figma documentation is enough. Maybe we should just make it a Sample? What do you think?

coraliefeil:
> @coraliefeil Thanks for the review :)
> 
> For the background story. You're right, It was implemented like a variant.
> 
> Since it is supposed to be a guideline only, I am now wondering if it should be a story at all and if the Figma documentation is enough. Maybe we should just make it a Sample? What do you think?

Yes, that’s a good idea. Could you do that, pls? THX

yoezlem (approved):

coraliefeil:
Approved. Thank you so much!

packages/components/src/components/accordion/accordion.ts
```diff
@@ -165,6 +164,12 @@ export default class SdAccordion extends SolidElement {
           @click=${this.handleSummaryClick}
           @keydown=${this.handleSummaryKeyDown}
         >
+          <div
+            class=${cx(
+              !this.open && 'opacity-0',
+              'w-1 bg-accent absolute left-0 transition-all h-[75%] group-hover:h-full'
```
mariohamann, line 170: I don't think that 75% is correct. I believe it should be more an absolute amount of pixels which is removed from the top and the bottom.

E. g.  if the full height is 100px → 75px, if the tull height is 200px → 175px. Otherwise it will stretch strangely in multiline buttons.

mariohamann (changes requested):

packages/components/src/components/accordion/accordion.ts
mariohamann: If this was implemented the same way in `sd-navigation-item` we should adapt it there, too.

mariohamann: But please clarify with design to make that sure.

Vahid1919:
@mariohamann @coraliefeil 
I updated the accordion's accent to use -16px of the total height instead of the -8px we discussed. This is double that of the sd-navigation-item, but it is pixel perfect with the design in Figma.

coraliefeil:
Approved. THX

mariohamann (approved):

PR #622, opened 2023-11-28
feat: ✨ add sd-radio-button

### LINKED ISSUE #216: feat: ✨ add sd-radio-button-group
## User Story
As a user of the Solid Design System, I need a Radio Button component that allows me to create button groups where only one button can be selected at a time. Additionally, I should have the option to display labels for each radio button in the group.

## API
https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=401-2292&mode=design

## Documentation
https://www.figma.com/file/T3mpP0ROquh93IDTqfFScU/Radio-Button-Group?type=design&node-id=0-1&mode=design

## Hand-off
https://www.figma.com/file/T3mpP0ROquh93IDTqfFScU/Radio-Button-Group?type=design&node-id=6-4&mode=design&t=ZUN1NQhEvk0PbgKZ-0

- [ ] Props, CSS-Properties, Parts & Slots
  - [ ] Capability to create exclusive button groups where only one button can be selected.
  - [ ] Option to display labels for each radio button.
- [ ] Stories
  - [ ] TBD

## DoR (Definition of Ready)
- [x] This user story provides value to the product.
- [x] The development team has estimated this user story.
- [x] The user story is clear and well-defined.
- [x] Any dependencies related to this user story have been identified.

## DoD (Definition of Done)
- [x] Documentation for the Radio Button component has been created or updated if applicable.
- [x] A migration guide is available or updated if applicable.
- [x] Relevant end-to-end tests for features, accessibility, and bug fixes are created or updated.
- [x] Relevant stories for features and accessibility are created or updated.
- [ ] Implementation successfully works on the `feature` branch.

azraefendic:
Added sd-radio-button component and radio-group example. Closes #216 

DoD (Definition of Done)
- [x] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [x] Implementation works successfully on feature branch


packages/components/src/components/radio-button/radio-button.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+    });
+  }
+};
+
+/**
+ * Use the show-label attribute to show the label of the radio button.
+ */
+
+export const CombinationAndSize = {
+  name: 'Combination × Size',
```
karlbaumhauer, line 89: As this story describes the `showLabel` attribute, I would like to see the story naming describing it more clear.

karlbaumhauer:
```suggestion
  name: 'Label × Size',
```

packages/components/src/components/radio-button/radio-button.stories.ts
```diff
@@ -0,0 +1,190 @@
...
+      args
+    });
+  }
+};
+
+/**
+ * Use the show-label attribute to show the label of the radio button.
+ */
+
+export const CombinationAndSize = {
```
karlbaumhauer, line 88:
```suggestion
export const LabelAndSize = {
```

packages/components/src/components/radio-button/radio-button.stories.ts
karlbaumhauer: or something like 'Additional Label x Size'

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+ * @status stable
+ * @since 1.0
+ *
+ * @slot - The radio button's label.
+ * @slot icon - A presentational icon.
+ *
+ * @event sd-blur - Emitted when the button loses focus.
+ * @event sd-focus - Emitted when the button gains focus.
+ *
+ * @csspart base - The component's base wrapper.
```
karlbaumhauer, line 23: How do I set the border radius to 0 if I have multiple buttons? Did not get it to work...

packages/components/src/components/radio-button/radio-button.stories.ts
```diff
@@ -0,0 +1,190 @@
+import '../../solid-components';
```
karlbaumhauer, line 1: I know the ticket description is not what we want to have but wouldnt it make sense to have a story with multiple buttons to show the functionality. Just by looking at it, I did not directly understand the use-case for this component.

This would also support the question on how to set the border radius to 0 when I have multiple buttons...

karlbaumhauer: ah... just found the story in the radio-group...

What is the reason, why we have it there? just to understand it...

karlbaumhauer (changes requested):

coraliefeil:
Review done. See Chromatic comments. Thanks a lot <3

Pls reassign for second review.

packages/components/src/components/radio-button/radio-button.stories.ts
abudd1094: I also find the presentation a bit strange, Shoelace uses groups in every example:
https://shoelace.style/components/radio-button

abudd1094:
In general, requirements are met and the component looks good!  I would prefer a a group of radio buttons in the examples like Shoelace, but this won't really work in the axis format we use.

The concept of stories could be refined at a higher level @karlbaumhauer @mariohamann .  Is the idea of combination stories like "disabled x size" that we have a visual "test" or check for every possible state for chromatic?  On the one hand this makes sense and on the other it convolutes our Storybook for users.  It seems to me that you can't have it both ways and that Storybook is best for documentation, not testing.

New contributors will be confused at the current state of our stories and they are starting to become divergent in nature.  It would be optimal to come together in the thought process.  That said, the concept for stories is left open to the developer at the moment, so in this case it is up to you @azraefendic .  I would prefer having one story for disabled and one for size.  The inclusion of three stories that show size feels redundant but makes sense from the testing perspective.

abudd1094 (approved):

karlbaumhauer:
> In general, requirements are met and the component looks good! I would prefer a a group of radio buttons in the examples like Shoelace, but this won't really work in the axis format we use.
> 
> The concept of stories could be refined at a higher level @karlbaumhauer @mariohamann @yoezlem . Is the idea of combination stories like "disabled x size" that we have a visual "test" or check for every possible state for chromatic? On the one hand this makes sense and on the other it convolutes our Storybook for users. It seems to me that you can't have it both ways and that Storybook is best for documentation, not testing.
> 
> New contributors will be confused at the current state of our stories and they are starting to become divergent in nature. It would be optimal to come together in the thought process. That said, the concept for stories is left open to the developer at the moment, so in this case it is up to you @azraefendic . I would prefer having one story for disabled and one for size. The inclusion of three stories that show size feels redundant but makes sense from the testing perspective.

@abudd1094 your mentioning a fair point and I understand the confusion. Thx for raising this point.
Our main point is to use the stories as visual tests with chromatic. In order to save snapshots, we started to combine stories. 
From my personal viewpoint, the docs file is the more interesting part for users. The stories are more to see them in action and prevent scrolling through the whole documentation. 
Another use-case for the stories are our "sample" stories, where we explicitly show a sample. 

This does not make it better but I am not sure how to fix this "confusion". If your willing, pls open up another ticket and write a suggestion. For now I would leave it like it is and then have another ticket to rework the whole library at once. This would then be a bigger new feature in the library but not mixing up the concepts in different components.

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+import { watch } from '../../internal/watch';
+import componentStyles from '../../styles/component.styles';
+import cx from 'classix';
+import SolidElement from '../../internal/solid-element';
+
+/**
+ * @summary Radio buttons allow the user to select a single option from a group using a button-like control.
+ * @documentation https://solid.union-investment.com/[storybook-link]/radio-button
+ * @status stable
+ * @since 1.0
```
yoezlem, line 15: Please keep in mind to change the version before merging! :)

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+  @property({ type: Boolean, reflect: true }) checked = false;
+
+  /** The radio's value. When selected, the radio group will receive this value. */
+  @property() value: string;
+
+  /** Disables the radio button. */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  /** The radio button's size. */
+  @property({ reflect: true }) size: 'sm' | 'md' | 'lg' = 'lg';
```
yoezlem, line 51: Please start first with the biggest size: 'lg' | 'md' | 'sm' = 'lg'

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  /** Shows or hides the label */
+  @property({ type: Boolean }) showLabel = false;
```
yoezlem, line 54: I have concern regarding the inconsistency in naming; one property is labeled as 'showLabel,' while the other is named 'hideIcon.' Could we standardize this and use either 'show' for both or 'hide' for both?

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+          @click=${this.handleClick}
+        >
+          <slot
+            name="icon"
+            part="icon"
+            class="${cx(
+              this.hasSlotController.test('icon') && 'inline-flex relative items-center',
+              this.size === 'sm' && 'text-base',
+              this.size === 'md' && 'text-lg',
+              this.size === 'lg' && 'text-xl'
```
yoezlem, line 139:
```suggestion
              {
                sm: 'text-base',
                md: 'text-lg',
                lg: 'text-xl'
              }[this.size]
```

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,236 @@
...
+      }
+
+      .sm-label {
+        height: 32px;
+      }
+
+      .sm-no-label {
+        height: 32px;
+        width: 32px;
+      }
```
yoezlem, line 197: Couldn't find the classes in html.. where is this being used?

yoezlem:
I am missing a story with more than one sd-radio-button.. for me as a user I wouldn't understand that. Or will this be another component (e.g. sd-radio-button-group)?

Also what about that info:
image

Do we need to add this to our component as a hint?

packages/components/src/components/radio-button/radio-button.stories.ts
azraefendic: I hear you, after trying out some different options I think it still makes sense to have examples with just one radio-button so that we can actually show the component and it's controls in the stories (we did the same with sd-radio for example). However I added more examples of button groups (both in sd-radio-button and sd-radio-group) and added more descriptions in documentation so it should all be more clear now I think. Let me know how it looks like to you.

packages/components/src/components/radio-button/radio-button.ts
azraefendic: I see your point, however it is also good practice to have the boolean props set to false by default? This was the reason for different naming (the icon is there by default and the label is not), but if it remains an issue maybe there's another way to approach it..

karlbaumhauer: @azraefendic your right (IMO). As defaults have to be false in a boolean attribute, this would be correct. 

in the radio-group we have:
```
  /**
   * The radio group's label. Required for proper accessibility. If you need to display HTML, use the `label` slot
   * instead.
   */
  @property() label = '';
```

Could we do something similar here as well?

packages/components/src/components/radio-button/radio-button.stories.ts
azraefendic: Well it technically is part of the sd-radio-group, I later added more examples to the sd-radio-button to make it more clear for users since I got feedback that it's not very intuitive and added different examples with and without labels so it's not repetitive, but I'm open to discuss this more and maybe find a better solution.. 

As for the mouseless story, similar topic, there is already one like that added in the sd-radio-group, so I'm thinking maybe there's no need for the first one to even exist? I see the same case in sd-radio..

karlbaumhauer: as discussed in our call:
- remove the group and mouseless story in the `sd-radio-button` component in order to be aligned with the `sd-radio`. 
- add a description to the `sd-radio-button` component which explains that its an equivalent to the `sd-radio` and its basically just a different radio-group-item next to the `sd-radio`.
- add the backgrounds from the current story (Groups and Backgrounds) to another story in the `sd-radio-button` to keep the backgrounds visible in the stories.
- remove or refactor the description of the radio-button-group stories in the `sd-radio-group` component in oder to make clear, that there is no such component as `sd-radio-button-group`

thx

coraliefeil:
Approved. Thanks a lot.

packages/components/src/components/radio-button/radio-button.ts
mariohamann: 1. Yah, or setting a `label` property and the slot overrides it. I think that would be best DX to have it similar. And if there's nothing, it should be hidden by itself.

2. Yes, but that's the same for every button which only shows an icon.

mariohamann (changes requested):

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,238 @@
...
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  /** The radio button's size. */
+  @property({ reflect: true }) size: 'lg' | 'md' | 'sm' = 'lg';
+
+  /** Shows or hides the label */
+  @property({ type: Boolean }) showLabel = false;
+
+  /** Hides the icon when set to true */
+  @property({ type: Boolean }) hideIcon = false;
```
mariohamann, line 57: I would remove this one as well and just watch the slot.

mariohamann (changes requested):

coraliefeil:
Approved.

packages/components/src/components/radio-button/radio-button.stories.ts
```diff
@@ -1,13 +1,16 @@
...
 import { storybookDefaults, storybookHelpers, storybookTemplate } from '../../../scripts/storybook/helper';
-import { waitUntil } from '@open-wc/testing-helpers';
 import { withActions } from '@storybook/addon-actions/decorator';
 
 const { argTypes, parameters } = storybookDefaults('sd-radio-button');
 const { overrideArgs } = storybookHelpers('sd-radio-button');
 const { generateTemplate } = storybookTemplate('sd-radio-button');
 
+/**
+ * The sd-radio-button is a type of a radio-group-item, same as the sd-radio. Radio button provides a button-style control for users to select one option from a group. They are specifically designed for integration within radio groups.
```
karlbaumhauer, line 11:
```suggestion
 * The `sd-radio-button` is a type of a radio-group-item, same as the `sd-radio`. Radio button provides a button-style control for users to select one option from a group. They are specifically designed for integration within radio groups.
```

karlbaumhauer (approved):
just one last suggestion in terms of formatting. Rest LGTM.

THX a lot for this component.

packages/components/src/components/radio-button/radio-button.ts
```diff
@@ -0,0 +1,234 @@
...
+
+      :host(.sd-button-group__button--first:not(.sd-button-group__button--last)) button {
+        border-start-end-radius: 0;
+        border-end-end-radius: 0;
+      }
+
+      :host(.sd-button-group__button--inner) button {
+        border-radius: 0;
+      }
+
```
Vahid1919, line 205: We can use Tailwind classes here now since Mario merged his PR yesterday. See https://github.com/solid-design-system/solid/pull/653

yoezlem (approved):

Vahid1919 (approved):

Issue #258, opened 2023-07-26
feat: ✨ add  sd-select


mariohamann:
## User Story
As a user of the Solid Design System, I require the `sd-select` component to facilitate the selection of items from a dropdown menu. This component should provide the following functionalities:

It will work like an original `<input type="select">`:

```
<sd-select label="Select one">
  <sd-option value="option-1">Option 1</sd-option>
  <sd-option value="option-2">Option 2</sl-option>
  <sd-option value="option-3">Option 3</sd-option>
</sl-select>
```

- [x] Display a dropdown menu containing a list of choices when clicked.
- [x] Keep the dropdown menu open until a menu item is selected or the user clicks outside of it.
- [x] In the `multiple` variant it should stay open until outside click
- [x] Allow hiding the default label for improved aesthetics.
- [x] Provide the option to display an additional hint for user guidance.
- [x] Automatically add a blank & an asterisk (*) to the label for mandatory fields to indicate their importance.
- [x] It should support multiple select like here: https://shoelace.style/components/select#multiple
- [x] In multiple select we could try to re-use the `sd-tag` and just override rectangle, as this could fit to the design: https://www.figma.com/file/ZphyVFsUHL72voMrJagMeo/Select?type=design&node-id=1013-8030&mode=design&t=8uInNVKl2s9IQzhf-0

By implementing these features, I can create user-friendly forms and interfaces that enable users to make selections efficiently, enhancing the overall usability of the system.

## Open questions
- [ ] can we use `sd-dropdown` for the select wrapper? Especially in terms of outline, sizes and status colors

## API

From design the following props are defined:

https://www.figma.com/file/ZphyVFsUHL72voMrJagMeo/Select?node-id=1002%3A2284&mode=dev

Besides those design related props we should add technical props of Shoelace: https://shoelace.style/components/select

## Hand-off
https://www.figma.com/file/ZphyVFsUHL72voMrJagMeo/Select?node-id=6%3A4&mode=dev

## Technical Information
Pls try to re-use the styles from #257 if possible.

### Components detailed requirements

We're building on https://github.com/solid-design-system/solid/blob/main/packages/components/src/_components/select/select.ts

We need `sd-select` and `sd-option`.
- [ ] Props, CSS-Properties, Parts & Slots
  - [ ] Implement props for configuring the behavior of `sd-select` components, such as label, hint, and mandatory field indicators.
  - [ ] Define CSS properties to allow customization of the component's appearance, including dropdown menu style and layout.
[...]

Issue #631, opened 2023-11-29
fix: 🤔 optimize initial validation in form components


karlbaumhauer:
>NOTE: 
This applies to all form elements / input components. The description below only handles the `sd-input` component as an example.
The fix should be implemented into all necessary components (`sd-switch`, `sd-checkbox`, `sd-radio-group`, `sd-input`, `sd-textarea` & `sd-select`).

## Current behavior
1. When using a required `sd-input` without any user interaction, the field is marked as red. This isn't very friendly UX: https://codepen.io/mariohamann/pen/KKJxXQx
2. When using a required `sd-input` with a prefilled value, the input still shows an error state (red). Only when interacting with the input, the components switches to the success state (green): https://codepen.io/mariohamann/pen/poGOWdO

## Expected behavior
There should be no validation indication until the user interacted with the field. It should never show an invalid/valid state as long as the user didn't interact.

## Technical Information
See https://shoelace.style/getting-started/form-controls#styling-invalid-form-controls for more details about the expected behavior. Especially about the attributes "data-valid" and "data-user-valid" to be used for visual validation states.

## Use-case:
The user builds a contact form, where the (logged in) end-user data is prefilled. The end-user then can edit the values or leave them like they are.
## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

-------------------------------------------------------------------------------------------------------------------------
As of 05.12.2023 we are combining the following ticket with this one: 
https://github.com/solid-design-system/solid/issues/501

We will use the `data-user-valid` and `data-user-invalid` attributes provided by the `formController` to condition our component styles based on form validity state and user interaction.

## Requirements: for each listed component...

- [ ] add private boolean state `showInvalidStyle`
- [ ] add private boolean state `showValidStyle` (if component has valid style)
- [ ] condition validity styles on `this.showValidStyle` and `this.showInvalidStyle`
- [ ] render div at bottom with id "error-message" for displaying inline error
- [ ] add query errorMessage for consistent access
- [ ] update handleInvalid method to display browser validation messages in errorMessage.textContent property

## Components
[...]

PR #634, opened 2023-12-01
feat: ✨ add sd-select #258

abudd1094:
Ready for initial review. Migration guide is in the works... it is unfortunately massive.


coraliefeil:
Design review done. Pls reassign for another review. Thanks a lot!

coraliefeil:
Pls reassign when adjusted. Thank you.

packages/components/src/components/option/option.ts
```diff
@@ -0,0 +1,195 @@
...
+            !this.disabled && this.hasHover ? 'h-full top-0' : 'top-1'
+          )}
+        ></span>
+        ${this.checkbox
+          ? html`<span
+              id="control"
+              part="control ${this.selected ? ' control--checked' : 'control--unchecked'}"
+              class=${cx(
+                'relative flex flex-initial items-center justify-center border rounded-sm h-4 w-4 mr-2',
+                this.disabled ? 'border-neutral-500' : this.selected ? 'bg-accent border-accent' : 'border-neutral-800'
```
yoezlem, line 150: To improve readability, I would recommend the following:

```suggestion
                this.disabled 
                  ? 'border-neutral-500' 
                  : this.selected 
                    ? 'bg-accent border-accent' 
                    : 'border-neutral-800'
```

packages/components/src/components/select/select.ts
```diff
@@ -47,109 +48,103 @@ import type SdPopup from '../popup/popup';
...
-  /** The default value of the form control. Primarily used for resetting the form control. */
-  @defaultValue() defaultValue: string | string[] = '';
+  @property({ type: Boolean, reflect: true }) open = false;
 
   /** The select's size. */
-  @property() size: 'small' | 'medium' | 'large' = 'medium';
+  @property({ reflect: true }) size: 'lg' | 'md' | 'sm' = 'lg';
+
+  /** The select's label. If you need to display HTML, use the `label` slot instead. */
+  @property() label = '';
```
yoezlem, line 112: We should think about a default value, cause label is (as I know) required for accessiblity 🤔

abudd1094: sure, does "Label" work?

packages/components/src/components/select/select.ts
```diff
@@ -629,53 +760,105 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
...
         part="form-control"
-        class=${classMap({
-      'form-control': true,
-      'form-control--small': this.size === 'small',
-      'form-control--medium': this.size === 'medium',
-      'form-control--large': this.size === 'large',
-      'form-control--has-label': hasLabel,
-      'form-control--has-help-text': hasHelpText
-    })}
+        class=${cx('form-control relative text-left', cursorStyles, textColor, textSize, this.open && 'z-10')}
```
yoezlem, line 831: If you are using the textColor only once, I would suggest to put it directly in the class (for better readability)

```
{
  disabled: 'text-neutral-500',
  readonly: 'text-black',
  activeInvalid: 'text-error',
  activeValid: 'text-success',
  active: 'text-black',
  invalid: 'text-error',
  valid: 'text-success',
  default: 'text-black'
}[selectState];
```

packages/components/src/components/select/select.ts
```diff
@@ -629,53 +760,105 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
...
-          <slot name="label">${this.label}</slot>
+          <slot name="label">${this.label}${this.required ? ' *' : ''}</slot>
         </label>
 
-        <div part="form-control-input" class="form-control-input">
+        <div part="form-control-input" class=${cx('relative w-full bg-white')}>
+          <div
+            part="border"
+            class=${cx(
+              borderColor,
```
yoezlem, line 847: Also here:

```
{
  disabled: 'border-neutral-500',
  readonly: 'border-neutral-800',
  activeInvalid: 'border-error border-2',
  activeValid: 'border-success border-2',
  active: 'border-primary border-2',
  invalid: 'border-error',
  valid: 'border-success',
  default: 'border-neutral-800'
}[selectState];
```

packages/components/src/components/select/select.ts
yoezlem: Also here:
```suggestion
        class=${cx(
          'form-control relative text-left', cursorStyles, this.open && 'z-10',
          this.disabled ? 'cursor-not-allowed' : 'cursor-pointer',
          this.size === 'sm' ? 'text-sm' : 'text-base',
          {
            disabled: 'text-neutral-500',
            readonly: 'text-black',
            activeInvalid: 'text-error',
            activeValid: 'text-success',
            active: 'text-black',
            invalid: 'text-error',
            valid: 'text-success',
            default: 'text-black'
          }[selectState]
        )}
```

yoezlem: It's just my opinion, but I think it's easier to read this way.

packages/components/src/components/select/select.ts
```diff
@@ -629,53 +760,105 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
...
         <label
           id="label"
           part="form-control-label"
-          class="form-control__label"
+          class=${hasLabel && 'inline-block mb-2'}
           aria-hidden=${hasLabel ? 'false' : 'true'}
           @click=${this.handleLabelClick}
         >
-          <slot name="label">${this.label}</slot>
+          <slot name="label">${this.label}${this.required ? ' *' : ''}</slot>
```
yoezlem, line 840: We should not handle different logics here. Please double-check whether we incorporate the 'required (*)' in the CSS or HTML and make the change consistently. Currently, it is displayed in the CSS for sd-radio, sd-switch, and sd-checkbox and in HTML for sd-input and sd-textarea.

packages/components/src/components/select/select.ts
```diff
@@ -629,53 +760,105 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
...
+    const borderColor = {
+      disabled: 'border-neutral-500',
+      readonly: 'border-neutral-800',
+      activeInvalid: 'border-error border-2',
+      activeValid: 'border-success border-2',
+      active: 'border-primary border-2',
+      invalid: 'border-error',
+      valid: 'border-success',
+      default: 'border-neutral-800'
+    }[selectState];
```
yoezlem, line 818: Then you could delete these lines.

yoezlem (changes requested):

packages/components/src/components/select/select.ts
yoezlem, line 818: Then you could removed these lines.

yoezlem:
Please assign me again when you need my review 🙂

packages/components/src/components/select/select.test.ts
```diff
@@ -64,10 +65,11 @@ describe('<sd-select>', () => {
         <sd-option value="option-2" disabled>Option 2</sd-option>
       </sd-select>
     `);
-    const disabledOption = el.querySelector('sd-option[disabled]')!;
+    const disabledOption: HTMLOptionElement = el.querySelector('sd-option[disabled]')!;
 
     await el.show();
     await clickOnElement(disabledOption);
+    // disabledOption.click();
```
karlbaumhauer, line 72: still needed?

packages/components/src/components/select/select.stories.ts
```diff
@@ -0,0 +1,427 @@
...
+  'tag',
+  'tag__base',
+  'tag__content',
+  'tag__remove-button',
+  'tag__remove-button__base',
+  'clear-button',
+  'expand-icon'
+];
+
+export const Parts = {
```
karlbaumhauer, line 242: I was wondering, if we need the stories for the parts `display-input` and `listbox`. From the description of the parts, these are not visible and thats probably the reason, why I cannot see any red box around it in Storybook.

What was your idea behind it? As a "new" user looking at it, it feels like there is something broken here. All other parts have a red outline but these 2... What do you think?

karlbaumhauer: sorry... 

same for `tag__remove-button__base`, `tag__remove-button` & `tag__content`

packages/components/src/components/select/select.stories.ts
```diff
@@ -0,0 +1,427 @@
...
+      })}
+    </div>`;
+  }
+};
+
+/**
+ * To allow multiple options to be selected, use the `multiple` attribute. It’s a good practice to use `clearable` when this option is enabled. To use the checkbox with tags variant, set the `useTags` variant to `true`.  To set multiple values at once, set value to a space-delimited list of values.  The preferred placement of the select’s listbox can be set with the `placement` attribute. Note that the actual position may vary to ensure the panel remains in the viewport. Valid placements are `top` and `bottom`.
+ */
+
+export const DisabledMultiple = {
```
karlbaumhauer, line 119: I am not sure if this is a bug in the `sd-tag` of here but if the `sd-select` is disabled, we still have a hover effect on the selected option. This seems wrong. Or is this the expected design?

packages/components/src/components/select/select.stories.ts
```diff
@@ -0,0 +1,427 @@
...
+      })}
+    </div>`;
+  }
+};
+
+/**
+ * Demonstrates the form behavior with validation styles when the `required` attribute is set to `true`.
+ */
+
+export const SampleForm = {
```
karlbaumhauer, line 354: I am not sure if this is a storybook/Chromatic bug but the validation still seems to have some issues here (also not sure if this is tackled in the validation ticket #631).

To reproduce:
1. without filling in any data, click submit => the browser default error message pops up on the first input and the first input is marked red  ✅
2. All other inoput fields do not change anything until you hover over them and they turn red as well ❓
3. Select an option in the first one and click submit again (without hovering over the others) => all others turn red as well  ✅

packages/components/src/docs/Migration/ui-select.mdx
```diff
@@ -0,0 +1,242 @@
...
+7. success
+8. variant
+
+<hr />
+
+## ✍️ CSS Variables
+
+### ✨ New CSS Variables
+
+#### All
```
karlbaumhauer, line 104: 😍 love that one... ;-)

packages/components/src/docs/Migration/ui-select.mdx
```diff
@@ -0,0 +1,242 @@
+# Migration Guide: From `ui-select` to `sd-select`
+
+The new `sd-select` is designed to replace the `ui-select`. Instead of mainly providing content via attributes, the `sd-select` component uses slots to allow for more flexibility and customization.
+
+## 💾 Slots
+
+### ✨ New Slots
+
+#### label
```
karlbaumhauer, line 9:
```suggestion
#### [label]
```

karlbaumhauer: pls change this in all level 4 headings as in the other migration guides. This improves readability and follows the structure of the other guides in this chapter.

karlbaumhauer (changes requested):
found some points to be clarified and small adjustments. 

Besides this: awesome work!!! Really nice and solid component / documentation. Especially the descriptions are super good to read and the docs presenting the component pretty good.

coraliefeil:
Found two small styling things ... pls reassign for final review. Thanks a lot!

packages/components/src/components/select/select.stories.ts
abudd1094: Good point, it isn't my idea so much as Shoelace's.  In my opinion, there are too many parts and they feel redundant like display-input and combobox, but do have different elements in the code.  I just kept the parts Shoelace has.  

Regarding the missing outlines, I don't understand why they aren't visible.  listbox for example has the correct style set but nothing is visible.  border on the is clearly visible:
image

I'm not sure how to fix this.  Any ideas @mariohamann ?

packages/components/src/components/select/select.stories.ts
abudd1094: this is indeed wrong

packages/components/src/components/select/select.stories.ts
abudd1094: @mariohamann aha, I guess this is why I was using that boolean property "enableValidation" that gets updated in the "updated" lifecycle method.  Without it, the component does not rerender until a hover interaction, at which point it applies the "isValid" or "isInvalid" styles... to fix this we need to add an updated lifecycle method for all form elements to update their validation states as soon as "data-user-valid" or invalid shows up on the element.  It could also be done with a mutation observer, but updated is easier... any thoughts?

abudd1094: My solution: each form element uses private states instead of constants in the render method for isValid and isInvalid.  The updated lifecycle method is used to react to attribute changes and update the corresponding state which will trigger an instant re-render.  Less code and easier to understand than a full blown mutation observer imo.

`  /**
   * Indicates whether or not the user input is valid after the user has interacted with the component. These states are activated when the attribute "data-user-valid" or "data-user-invalid" are set on the component via the form controller.
   */
  @state() private isValid = false;
  @state() private isInvalid = false;
  
 /** Checks for the presence of the attributes 'data-user-valid' or 'data-user-invalid' and updates the corresponding state for reactive conditional styling. */
  updated() {
    if (this.hasAttribute('data-user-valid') && this.checkValidity()) {
      this.isValid = true;
    }

    if (this.hasAttribute('data-user-invalid') && !this.checkValidity()) {
      this.isInvalid = true;
    }
  }
  `

packages/components/src/components/select/select.stories.ts
abudd1094: Ok, the other issues are resolved, this is the last standing topic.  I can take a closer look but I'm not sure what the best solution is here.

coraliefeil:
Approved. Merci <3

packages/components/src/components/select/select.ts
```diff
@@ -773,21 +789,18 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
     const hasHelpText = this.helpText ? true : !!slots['helpText'];
     const hasClearIcon = this.clearable && !this.disabled && this.value.length > 0;
 
-    const isInvalid = this.hasAttribute('data-user-invalid') && this.required && !this.checkValidity();
-    const isValid = this.hasAttribute('data-user-valid') && this.required && this.checkValidity();
-
     // Hierarchy of input states:
     const selectState = this.disabled
```
karlbaumhauer, line 793: 🤯

karlbaumhauer: love the query...;-)

packages/components/src/components/select/select.stories.ts
karlbaumhauer: @abudd1094 pls open up a bug ticket to tackle the issue. I agree to the fact, that we should at least question the parts and/or make them be visible here.

karlbaumhauer (approved):
LGTM. Awesome new component.

PR #637, opened 2023-12-04
feat: ✨ add sd-chip

### LINKED ISSUE #272: feat: ✨ add sd-chip (CSS style component)
## User Story
As a user of the Solid Design System, I would like to have a `sd-chip` component, a small, non-interactive label that represents a status, property, or other metadata

## Hand-off
https://www.figma.com/file/IUiRoK2jiW8ydM77uiY2RX/Chip?type=design&node-id=1103-913&mode=design&t=9jVZgnCcpvwX9rLA-4
## Documentation
https://www.figma.com/file/IUiRoK2jiW8ydM77uiY2RX/Chip?type=design&node-id=804-717&mode=design&t=9jVZgnCcpvwX9rLA-4

#### Props
| <ul><li>[ ] </li></ul>  | `--variant-` | `white` / `primary--200` / `primary--300` / `primary--500` | `primary--200` | defines the background color |

  - [x] variant (shows different color variants and in which combination it should be used)

#### Samples
- ~~A sample that shows and highlights the difference between `sd-flag` & `sd-chip` => import from #275~~

## DoR
- [x] Item has business value

## DoD
- [x] Implementation works successfully on `feature` branch

saemik94:
## Description:
Added sd-chip style. Closes #272


coraliefeil:
Review done. Pls reassign for second review. Thanks.

yoezlem:
Would it make sense to create a sample story like in figma?
image

karlbaumhauer:
> Would it make sense to create a sample story like in figma? 

its mentioned in the ticket, that the story will be added to the sd-flag, as it shows the different use-cases.

saemik94:
> The default Variant should always be at top: image

The sort order is numbers and then strings, all currents examples are strings like '3xl'  an so on, therefore the sort order was the right way - but this is just luck. When you have numbers in the list, they will be put in front... But I had to adjust the default value and due to that now also have string options -> the sort oder is the right way again

abudd1094 (approved):

yoezlem (approved):

PR #663, opened 2023-12-19
fix: 🤔 teaser variant 'white border-neutral-400', - no border around media slot

azraefendic:
## Description:
Added negative margin on the media slot in the teaser variant 'white border-neutral-400' so that it overlaps the border and matches the design.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/teaser/teaser.ts
+++ b/packages/components/src/components/teaser/teaser.ts
@@ -110,7 +110,7 @@ export default class SdTeaser extends SolidElement {
           class=${cx(
             !inset && this._orientation === 'vertical' && 'mb-4',
             !slots['teaser-has-media'] && 'hidden',
-            this.variant === 'white border-neutral-400' && '-m-px'
+            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && '-m-px'
           )}
           part="media"
         >
--- a/packages/components/src/components/teaser/teaser.ts
+++ b/packages/components/src/components/teaser/teaser.ts
@@ -110,7 +110,7 @@ export default class SdTeaser extends SolidElement {
           class=${cx(
             !inset && this._orientation === 'vertical' && 'mb-4',
             !slots['teaser-has-media'] && 'hidden',
-            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && '-m-px'
+            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && 'm-[-1px]'
           )}
           part="media"
         >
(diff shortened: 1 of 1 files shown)


packages/components/src/components/teaser/teaser.ts
```diff
@@ -107,7 +107,11 @@ export default class SdTeaser extends SolidElement {
       >
         <div
           style=${this._orientation === 'horizontal' ? `width: var(--distribution-media, 100%);` : ''}
-          class=${cx(!inset && this._orientation === 'vertical' && 'mb-4', !slots['teaser-has-media'] && 'hidden')}
+          class=${cx(
+            !inset && this._orientation === 'vertical' && 'mb-4',
+            !slots['teaser-has-media'] && 'hidden',
+            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && '-m-px'
```
Vahid1919, line 113: The solution works great. I considered utilizing arbitrary values in this context though.

```suggestion
            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && 'm-[-1px]'
```

Vahid1919 (approved):

packages/components/src/components/teaser/teaser.ts
mariohamann: Why? Arbitrary values are last resort and should be avoided. If someone else uses -m-px somewhere else (what would be correct!) we have two classes for the same thing, what bloats our CSS.

mariohamann: Ah. Now I get the problem. We didn't hab"ve the class. :D What's a problem of our spacing scale... let's keep that in mind. We should add that to our scale.

Vahid1919: Created a ticket for it here -> [#664](https://github.com/solid-design-system/solid/issues/664)

packages/components/src/components/teaser/teaser.ts
```diff
@@ -107,7 +107,11 @@ export default class SdTeaser extends SolidElement {
       >
         <div
           style=${this._orientation === 'horizontal' ? `width: var(--distribution-media, 100%);` : ''}
-          class=${cx(!inset && this._orientation === 'vertical' && 'mb-4', !slots['teaser-has-media'] && 'hidden')}
+          class=${cx(
+            !inset && this._orientation === 'vertical' && 'mb-4',
+            !slots['teaser-has-media'] && 'hidden',
+            this.variant === 'white border-neutral-400' && this._orientation === 'vertical' && 'mx-[-1px] mt-[-1px]'
```
mariohamann, line 113: What is with orientation horizontal?

azraefendic: It always has inset set to true

mariohamann (approved):

PR #704, opened 2024-01-24
feat: ✨add shadow variant for sd-table-cell 

### LINKED ISSUE #372: feat: ✨ sd-table-cell add shadow property/style
## User Story
As a user of the Solid Design System, I would like to set the first column in a table to fixed using the sd-table component while keeping it fully accessible and according to the provided design (shadow).

### Suggesed Solution
As we don't provide any html table logic/elements with the sd-table component, the easiest and most maintainable way would be to provide a sd-table-cell property to add a shadow to the right side of the sd-table. 
In addition to this, through a `scrollable` story, we would provide a sample how to setup the logic inside the html table to make it scrollable.

## API
tba from https://github.com/solid-design-system/figma/issues/124

## Hand-off
tba from https://github.com/solid-design-system/figma/issues/124

- [ ] Props
  - [ ] `shadow`: left, right, top & bottom
  - [ ] `shadow-active`
- [ ] Stories
  - [ ] shadow
- [ ] Samples
  - [ ] `shadow` in several directions
  - [ ] `shadow` on small screens

## DoR
- [ ] Item has business value
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD

saemik94:
closes #372 

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/table-cell/table-cell.css
+++ b/packages/components/src/styles/table-cell/table-cell.css
@@ -11,9 +11,9 @@
       position: absolute;
       left: 0;
       right: 0;
-      bottom: -5px; /* Adjust the distance of the shadow from the header row */
-      height: 5px; /* Adjust the height of the shadow */
-      background: linear-gradient(to bottom, rgba(0, 0, 0, 0.2), transparent);
+      bottom: 0;
+      height: 10px;
+      background: linear-gradient(to bottom, rgba(24, 24, 24, 0.1), transparent);
     }
   }
 
@@ -24,8 +24,8 @@
       top: 0;
       bottom: 0;
       right: 0;
-      width: 5px; /* Adjust the width of the shadow as needed */
-      background: linear-gradient(to right, rgba(0, 0, 0, 0.2), transparent);
+      width: 10px; /* Adjust the width of the shadow as needed */
+      background: linear-gradient(to right, rgba(24, 24, 24, 0.1), transparent);
       border-style: none;
     }
   }
--- a/packages/components/src/styles/table-cell/table-cell.css
+++ b/packages/components/src/styles/table-cell/table-cell.css
@@ -7,7 +7,7 @@
 
   &--shadow-bottom {
     &:after {
-      @apply opacity-0 absolute pointer-events-none transition-opacity duration-300 left-0 right-0 h-2.5 bottom-[-10px] content-[''] bg-gradient-to-b from-neutral-300 to-transparent;
+      @apply opacity-0 absolute pointer-events-none transition-opacity duration-300 left-0 right-0 h-2.5 bottom-[-10px] content-[''] bg-gradient-to-b from-[#000]/10 to-transparent;
     }
 
     &.sd-table-cell--shadow-active:after {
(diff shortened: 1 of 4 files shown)


packages/components/src/styles/table-cell/table-cell.css
```diff
@@ -5,6 +5,32 @@
...
+  &--shadow-bottom {
+    &:after {
+      content: '';
+      position: absolute;
+      left: 0;
+      right: 0;
+      bottom: -10px;
+      height: 10px;
+      background: linear-gradient(to bottom, rgba(24, 24, 24, 0.1), transparent);
+      transition: opacity 10s linear;
```
mariohamann, line 17: 10 seconds? 😆 

And it feels that you could abstract some stuff.

&--shadow-bottom, &--shadow-right{

content: '';
transition...;
position: absolute;
}

...

So that we clearly see differnce.

I'm a bit irritated, that we need first-child for shadow-right? This should be part of the element directly?

The transition currenlty doesn't work. I believe you have to make the content and everythign visible and just change opacity etc.

Plus make sure to remove pointer-events, so no its clickthroughable.

mariohamann (changes requested):

packages/components/src/styles/table-cell/table-cell.css
mariohamann: Oh and please use Tailwind here:

@left-0 right-0 etc.

packages/components/src/styles/table-cell/table-cell.css
```diff
@@ -5,6 +5,56 @@
...
+
+    /* fix that shows line on top of table-cell */
+    &:before {
+      @apply bg-neutral-400 block content-[''] absolute top-[-1px] w-full h-[1px] left-0;
+    }
+  }
+
+  &--shadow-right {
+    &:after {
+      @apply opacity-0 absolute pointer-events-none transition-opacity duration-1000 content-[''] bg-gradient-to-r from-neutral-300 to-transparent top-0 bottom-0 right-[-10px] w-2.5;
```
karlbaumhauer, line 25: just out of curiosity: cant we combine those classes into a single class we can use for other shadows in other components? 

I see `opacity-0 absolute pointer-events-none transition-opacity duration-1000 from-neutral-300 to-transparent` as duplicates in here, which only describe the kind of shadow we want to have (not sure about the `absolute`).

Feels like we will have the same with other shadows or is this the only use case for a shadow?

packages/components/src/styles/table-cell/table-cell.css
```diff
@@ -5,6 +5,56 @@
     @apply border-r-[1px];
   }
 
+  &--shadow-bottom {
+    &:after {
+      @apply opacity-0 absolute pointer-events-none transition-opacity duration-1000 ease-out	left-0 right-0 h-2.5 bottom-[-10px] content-[''] bg-gradient-to-b from-neutral-300 to-transparent;
```
karlbaumhauer, line 10: why does this shadow have a different transition timing function (ease-out) than all other shadows?

saemik94: That shouldn't be the case, it's a remnant of development.

packages/components/src/styles/table-cell/table-cell.css
karlbaumhauer: @coraliefeil do you see an upcoming reuse of this type of transition?

saemik94: @mariohamann should we create a token for this, or how to deal with that?

coraliefeil:
Review done. Pls reassign for final check. Thanks a lot.

coraliefeil:
Approved from design perspective. Thank you so much!

packages/components/src/styles/table-cell/table-cell.css
saemik94: outdated

packages/components/src/styles/table-cell/table-cell.css
saemik94: refactored with the suggestions from @mariohamann

karlbaumhauer (approved):
LGTM, pls remember to adopt PR description before merging.

PR #788, opened 2024-02-28
fix: arbitrary values in container style

abudd1094:
## Description:
convert all tailwind arbitrary values to plain css in container style

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/container/container.declaration.ts
+++ b/packages/components/src/styles/container/container.declaration.ts
@@ -21,7 +21,7 @@ export default {
       name: 'sd-container--triangle-...',
       options: ['top', 'right', 'bottom', 'left'],
       description:
-        'Defines optional triangle cut-out for sd-container. This allows for an indentation resembling an arrow on any side of the container. CSS Property `triangle-background` defines the background color of the cut-out.'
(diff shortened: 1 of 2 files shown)


packages/components/src/styles/container/container.declaration.ts
```diff
@@ -21,7 +21,7 @@ export default {
       name: 'sd-container--triangle-...',
       options: ['top', 'right', 'bottom', 'left'],
       description:
-        'Defines an optional triangle cut-out for sd-container. This allows for an indentation resembling an arrow on any side of the container. CSS Property `triangle-background` defines the background color of the cut-out.'
+        'Defines optional triangle cut-out for sd-container. This allows for an indentation resembling an arrow on any side of the container. CSS Property `triangle-background` defines the background color of the cut-out.'
```
Vahid1919, line 24: Very minor but I think it makes sense to keep the "an". Approving since this is non-blocking.

```suggestion
        'Defines an optional triangle cut-out for sd-container. This allows for an indentation resembling an arrow on any side of the container. CSS Property `triangle-background` defines the background color of the cut-out.'
```

Vahid1919 (approved):

packages/components/src/styles/container/container.declaration.ts
abudd1094: Haha, I agree, I just needed the pipeline to run and for some reason merging main did not work...

PR #845, opened 2024-03-07
feat: ✨ add supernumber pattern 

### LINKED ISSUE #307: feat: ✨ add supernumber pattern
## User Story
As a Solid Design System user, I need a supernumber pattern to highlight key figures. It should include overline, number, subline, and description for clear and consistent presentation of important numerical data.

## Documentation
https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=1663-16479&mode=design&t=4DSpu9hffmhPXlqI-4

## Hand-off
https://www.figma.com/file/DAidYiZR80xAc8nr7eGI3m/Supernumber?type=design&node-id=6-4&mode=design&t=Nyyt2RmAX0zIJKia-0

  - [x] `Default` - for the text above, below & description
  - [x] `Prefix` - is the prefix inside the supernumber -> for (prefix) 200 (number) years (suffix)
  - [x] `Suffix` - is the suffix inside the supernumber
  - [x] `Size` - include 3 sizes, depending on the container-size -> lg, md, sm
  - [x] `Inverted`
  - [x] `i18n` - show internationalization with countUp.js options

### Technical Information
- [x] The text above and under the number are `sd-paragraph` with bold font-weight.
- [x] For the devs: don't use `sd-display` for the number. Adapt everything from the typography section in figma.
- [x] The description is `sd-paragraph`
- [x] The counter will be implemented with https://inorganik.github.io/countUp.js/

## DoR
- [x] Item has business value

## DoD
- [x] Implementation works successfully on `feature` branch

Vahid1919:
## Description:
Added supernumber pattern. Closes #307.

@karlbaumhauer if I am remembering correctly we decided to move away from using "sd-" for patterns. Therefore, I renamed this PR and made adjustments to the Migration guide to reflect that change. 

@solid-design-system/development  As you know, the default story renders twice in the Docs page. This was causing issues with countup.js so I used inline=false. This results in the story being unable to start countup **on scroll** like the remaining stories. I thought that is a reasonable trade off. But open to new ideas :) 

## Definition of Reviewable:


karlbaumhauer (changes requested):
nice work though...

packages/components/src/components/checkbox-group/checkbox-group.test.ts
```diff
@@ -14,8 +14,8 @@ describe('when submitting a form', () => {
       </form>
     `);
     const button = form.querySelector('sd-button')!;
-    const checkbox = form.querySelectorAll('sd-checkbox')[1]!;
-    const checkbox2 = form.querySelectorAll('sd-checkbox')[2]!;
+    const checkbox = form.querySelectorAll('sd-checkbox')[1];
```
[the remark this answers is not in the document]
Vahid1919: I added the change here just to avoid the error. When #709 is merged, this will disappear from the "files changed" as it will be the same in main.

karlbaumhauer:
> @solid-design-system/development As you know, the default story renders twice in the Docs page. This was causing issues with countup.js so I used inline=false. This results in the story being unable to start countup on scroll like the remaining stories. I thought that is a reasonable trade off. But open to new ideas :)

@Vahid1919 I think this solution is fine, I would just like to see a short note on that for the users.

packages/components/src/patterns/supernumber/supernumber.stories.ts
```diff
@@ -0,0 +1,272 @@
+import '../../solid-components';
```
karlbaumhauer, line 1: I don't know if this is related to your [comment](https://github.com/solid-design-system/solid/pull/845#issuecomment-1983852520) or to chromatic but I had the behavior, that after switching too fast between the stories, none of them BUT the default story started counting on scroll. I had the reload the page to get it working again. 

If its only with chromatic, I am fine but otherwise this should be fixed... let me know if you need a screen sharing session to reproduce it.

Vahid1919: This is a chromatic limitation/quirk. When navigating through stories, the page is not reloaded so when new stories are run it assumed we are redeclaring const/let variables. This causes the js to crash. 
image

This can be resolved by using the `var` keyword and scoping them globally. But that would not be ideal to put in a sample as `var` is not best practice. Especially since this issue is limited to Chromatic.

I am not sure whether we should use `var` but add a disclaimer or use const and still add a disclaimer.

packages/components/src/patterns/supernumber/supernumber.stories.ts
```diff
@@ -0,0 +1,272 @@
...
+      <div class="text-[72px] text-primary" id="countup">0</div>
+      <p class="sd-paragraph font-bold my-4">Nisi eu excepteur anim esse</p>
+
+      <p class="sd-pararaph sd-paragraph--size-sm text-center  pt-2">
+        Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmd tempor invit ut labore et dolore
+        magna aliquyam erat
+      </p>
+    </div>
+
+    <script src="https://inorganik.github.io/countUp.js/dist/countUp.umd.js"></script>
```
mariohamann, line 36: Please no external scripts.

packages/components/package.json
```diff
@@ -80,6 +80,7 @@
     "@shoelace-style/animations": "^1.1.0",
     "@shoelace-style/localize": "^3.1.2",
     "classix": "^2.1.35",
+    "countup.js": "^2.8.0",
```
mariohamann, line 83: Do not add this to dependencies, this is only for documentation and is therefore a devDependency.

packages/components/src/docs/Migration/ui-supernumber.mdx
```diff
@@ -0,0 +1,54 @@
...
+11. [text-before]
+12. [variant]
+13. [width]
+
+<hr />
+## 🥳 Events
+
+### ❌ Removed Events:
+
+The following events have been removed from the new `supernumber` pattern:
```
mariohamann, line 40: As a pattenr per design can't have those, I would just write a sentence, that therefore no attributes, events or methods are available.

mariohamann (changes requested):

coraliefeil:
@Vahid1919 good job!

Would like to make a bit more clear that the default is a static number and that as an option the number can be animated using countUp.js.

Use font size text-base for the description below, pls.

Check vertical padding between overline and number and underline. In code they are bigger though the values seem alright from inspecting:
Bildschirmfoto 2024-03-12 um 12 06 40

Vahid1919:
> @Vahid1919 good job!
> 
> Would like to make a bit more clear that the default is a static number and that as an option the number can be animated using countUp.js.
> 
> Use font size text-base for the description below, pls.
> 
> Check vertical padding between overline and number and underline. In code they are bigger though the values seem alright from inspecting: Bildschirmfoto 2024-03-12 um 12 06 40

@coraliefeil Thanks for the feedback. I noticed the line height was missing, that was the issue. 

There seems to be values here (size and line height), that are not in the tokens. Is there a way we can include these as tokens alongside this ticket?

Vahid1919:
@coraliefeil I reformatted the stories to make it more clear that the countupjs stuff is optional. Please take a look :)

karlbaumhauer (approved):
Approved concerning the change requests I had. Looks good.

Vahid1919:
@coraliefeil I made the design changes here. Please Review when you have time :)

coraliefeil:
Fantastic. All good. Merci <3

PR #847, opened 2024-03-08
fix: adjust vertical spacing of navigation-item#vertical

christophsaile:
## Description:
*PR notes: Please describe the changes in this PR.*

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/navigation-item/navigation-item.ts
+++ b/packages/components/src/components/navigation-item/navigation-item.ts
@@ -139,12 +139,11 @@ export default class SdNavigationItem extends SolidElement {
       <${tag}
         part="base"
         class=${cx(
-          'hover:bg-neutral-200 group transition-all min-h-[48px] cursor-pointer relative focus-visible:focus-outline',
+          'px-4 hover:bg-neutral-200 group transition-all min-h-[48px] cursor-pointer relative focus-visible:focus-outline',
           { base: 'text-base', lg: 'text-lg', sm: 'text-[14px]' }[this.size],
           this.disabled ? 'text-neutral-500 pointer-events-none' : 'text-primary',
           isAccordion ? 'flex flex-col' : 'inline-block w-full',
-          this.divider && this.vertical && 'mt-[1px]',
-          this.vertical ? 'px-8' : 'px-4'
+          this.divider && this.vertical && 'mt-[1px]'
         )}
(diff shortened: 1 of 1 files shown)


karlbaumhauer (approved):

christophsaile:
> Pls reassign for another review. THX

Hey @coraliefeil I can't reply to our review in chromatic so i answer here. I fixed the first point you mentioned
Regarding the second point
grafik
Shouldn't the padding bottom for the case false false also be adjusted otherwise I think if they are placed next to eachother the text will not be aligend.
@karlbaumhauer @mariohamann currently there is no token for padding-bottom: 10px should i add `pb-[0.625rem]`?

yoezlem (approved):

karlbaumhauer:
> > Pls reassign for another review. THX
> 
> Hey @coraliefeil I can't reply to our review in chromatic so i answer here. I fixed the first point you mentioned Regarding the second point grafik Shouldn't the padding bottom for the case false false also be adjusted otherwise I think if they are placed next to eachother the text will not be aligend. @karlbaumhauer @mariohamann currently there is no token for padding-bottom: 10px should i add `pb-[0.625rem]`?

@mariohamann / @Vahid1919 can you answer/support on the padding bottom question?

@coraliefeil / @EllaELSA do we actually have the use case @christophsaile mentioned?

christophsaile:
> > > > Pls reassign for another review. THX
> > > 
> > > 
> > > Hey @coraliefeil I can't reply to our review in chromatic so i answer here. I fixed the first point you mentioned Regarding the second point grafik Shouldn't the padding bottom for the case false false also be adjusted otherwise I think if they are placed next to eachother the text will not be aligend. @karlbaumhauer @mariohamann currently there is no token for padding-bottom: 10px should i add `pb-[0.625rem]`?
> > 
> > 
> > @mariohamann / @Vahid1919 can you answer/support on the padding bottom question?
> > @coraliefeil / @EllaELSA do we actually have the use case @christophsaile mentioned?
> 
> We try to avoid custom values as much as possible. We do have instances in the code where we provided them before when we lacked any other workaround or tokens can't/won't be provided. I would check if it makes sense to update the tokens to include these values (@solid-design-system/core-design ?) Otherwise, yes use the custom css values.

If this whole topic takes some time to discuss I would suggest to move the discussion about the padding between the text and and the bottom of the element to another ticket since this is currently blocking one of our sprint goal tickets.
Furthermore fixing this is not part of the ticket scope.

PR #875, opened 2024-03-19
docs: 📚 add migration guide from ui-image to sd-media and sd-copyright [skip chromatic]

### LINKED ISSUE #808: docs: 📚 add migration guide from `ui-image` to `sd-media` and `sd-copyright`
## Description
The `ui-image` has been transformed to `sd-media` and `sd-copyright`. Unfortunately a migration guide was not created and need to be subsequently submitted.

## Where to put the documentation
Storybook
## DoR
- [x] Item has business value

## DoD

azraefendic:
## Added a migration guide from ui-image to sd-media and sd-copyright

Closes #808 

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/docs/Migration/ui-image.mdx
+++ b/packages/components/src/docs/Migration/ui-image.mdx
@@ -1,18 +1,20 @@
 # Migration Guide: `ui-image`
 
-The ui-image has been transformed to `sd-media` and `sd-copyright`.
+The new `sd-media` and `sd-copyright` are designed to replace the `ui-image`. Rather than offering a complex component structure, they primarily focus on providing CSS styles.
 
-# ui-image to sd-media
+Media no longer requires the `ui-image` wrapper and instead can directly utilize the provided CSS classes.
 
-Instead of mainly providing content via attributes, the new `sd-media` component uses slots to allow for more flexibility and customization.
+## 🎨 CSS Classes
 
-## 💾 Slots
+### ✨ New CSS Classes
 
-### ✨ New Slots
+#### [sd-media]
 
-#### [default]
+The `sd-media` class is used to apply the basic styles to the media elements such as margin and default text color.
 
-Content of the element.
+#### [sd-copyright]
+
+The `sd-copyright` class is used to apply the basic styles to the copyright elements. It positions the element and defines padding, font size, color and text shadow.
 
 <hr />
 
@@ -20,7 +22,7 @@ Content of the element.
 
 ### ❌ Removed Attributes
 
-The following attributes have been removed from the new sd-media component:
+Since the new [sd-media] and [sd-copyright] are no longer components basically all attributes have been removed:
 
 1. [altText]
 2. [componentWidth]
@@ -48,7 +50,7 @@ The following attributes have been removed from the new sd-media component:
 
 ### ❌ Removed Events:
 
-The following events have been removed from the new sd-media component:
+The following events have been removed from the new components:
 
 1. [componentResized]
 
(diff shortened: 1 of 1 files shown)


packages/components/src/docs/Migration/ui-image.mdx
```diff
@@ -0,0 +1,127 @@
+# Migration Guide: `ui-image`
+
+The ui-image has been transformed to `sd-media` and `sd-copyright`.
+
+# ui-image to sd-media
+
+Instead of mainly providing content via attributes, the new `sd-media` component uses slots to allow for more flexibility and customization.
```
yoezlem, line 7: sd-media and sd-copyright are not using slot, they now are classes which can be used in HTML. Perhaps you could orient yourself according to the sd-table 🤔

packages/components/src/docs/Migration/ui-image.mdx
```diff
@@ -0,0 +1,67 @@
+# Migration Guide: `ui-image`
+
+The new `sd-media` and `sd-copyright` are designed to replace the `ui-image`. Rather than offering a complex component structure, they primarily focus on providing CSS styles.
```
mariohamann, line 3: Please make more clear, what the two components do and hwo they should/could be used. See e. g. https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/styles-sd-media--docs

It could be even worth to show some example code here. But do not feel forced to do so.

But definitely describe better, what each CSS does.

packages/components/src/docs/Migration/ui-image.mdx
```diff
@@ -0,0 +1,67 @@
...
+The following events have been removed from the new components:
+
+1. [componentResized]
+
+<hr />
+
+## 🧪 Methods
+
+### ❌ Removed Methods:
+
```
Vahid1919, line 62: I think we could remove the list of individual attributes since it doesn't make sense to have attributes in a CSS style in the first place.

We decided to not include these for pattern migration guides for example ([ui-suppernumber](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/docs-migration-ui-supernumber--docs)) I think it makes sense to do the same here.

azraefendic: I can see your point but this would create too many different guides, since we already used this structure for other CSS-styles migration guides.

mariohamann: I think both of you are right. I would follow our current approach, but at least have the concerns in mind, especially if complains are coming in on that. I'm hesitating doing a ticket for that, as it's maybe not worth the effort. But I would say that's the decision of @karlbaumhauer and @yoezlem

yoezlem:
Maybe thats something for our new next stage "Documentation" where we will add more documentation and stuff in our Storybook. So we can than take also a deeper look if we should change our migration guides for the styles and pattern.
I suggest we maintain the current status for now, and revisit it at a later time for further evaluation 👍🏻

Vahid1919 (approved):

packages/components/src/docs/Migration/ui-image.mdx
Vahid1919: I see. Considering the other CSS migration guides did include it, this definitely makes sense. Resolving this conversation to unblock this PR.

karlbaumhauer: From my point of view the removed attributes have a value here. 
1st, we removed them and it might help in one or another case to already provide a technical hint on what to use in return here (eg. a variant) and 2nd, we keep the guides in a similar style and they will be easier to read. 

It might (as 3rd point) even help users to summarize the functionality of the component and tell them exactly what they have to implement on their side from now on.

karlbaumhauer (approved):
besides what @mariohamann mentioned. LGTM. Thx.

PR #995, opened 2024-04-29
fix: exclude event names from versioning

mariohamann:
## Description:
This excludes event names from versioning. 

Before that the event `sd-input` was renamed to `sd-2-11-0-input`. This came from the fact, that the event had the same name as a component. This change omits this. 

## Definition of Reviewable:


yoezlem (approved):
LGTM

PR #997, opened 2024-04-29
feat: ✨ add custom localization (closes #852)

### LINKED ISSUE #852: feat: ✨ integrate instance specific i18n
## User Story
As a developer using Solid components, I want to be able to override baked-in strings of components on an instance-level.
 
### Current approach
We're having [translation files](https://github.com/solid-design-system/solid/blob/main/packages/components/src/translations/de.ts) in which we describe every translatable string, e. g.:

https://github.com/solid-design-system/solid/blob/55f27132938aa8e19b9e5e73e4d729ba7889af63/packages/components/src/translations/de.ts#L9

This is then picked up in a component, depending on `lang` attribute on body or the component, see e. g. https://github.com/solid-design-system/solid/blob/55f27132938aa8e19b9e5e73e4d729ba7889af63/packages/components/src/components/carousel/carousel.ts#L105

To make this overridable we're starting to add custom props to override translatable strings, e. g. https://github.com/solid-design-system/solid/issues/746 and https://github.com/solid-design-system/solid/issues/300. To make this approach complete, we would have to do this for every translatable single string on every component, which doesn't feel well.

### Suggested Solution
We could alter or extend the `shoelace-stylize/localize` package so that in addition to checking `lang` attributes it should watch out for the `data-custom-localization` attribute on component or body label. If this is set and there is a translation for a string in the component, it should be automatically picked up instead of the "fallback" in the translation files.

Example:

```html
<sd-carousel i18n='{"carousel":"Kah-Rau-Sel"}></sd-carousel>
```

or

```html
<sd-carousel custom-localize='{"carousel":"Kah-Rau-Sel"}></sd-carousel>
```

This approach would work for most translations. Still for dynamic translations which need a function, this would have to be set via JS:
```js
const el = document.querySelector("sd-select#selected-funds");;
el.i18n.numOptionsSelected = num => {
    if (num === 0) return '';
    return `${num} Fonds ausgewählt`;
  };
```

It has to be documented, that the later shouldn't be implemented that editors in a CMS can override this, as this is evaluated code which could be misused.

## ACs

- [ ] create localization docs page based on the docs from shoelcae and add information about custom localization to it
- [ ] apply code suggestions from the draft PR
- [ ] cleanup (if neccessary) code from the PoC #997 

## DoR
- [x] Item has business value

## DoD

mariohamann:
## Description:

This branch offers the option for every single element, to be translated on an instance level. Check the tests and docs for what it does.

Closes https://github.com/solid-design-system/solid/issues/852

## Definition of Reviewable:


packages/components/src/utilities/localize.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      <sd-select lang="de" value="option-1" clearable><sd-option value="option-1">Option 1</sd-option></sd-select>
+    `);
+    expect(select.shadowRoot!.querySelector('button[part="clear-button"]')!.getAttribute('aria-label')).to.equal(
+      'Eingabe löschen'
+    );
+  });
+
+  it('should apply custom localization from data-attribute', async () => {
+    const select = await fixture<SdSelect>(html`
+      <sd-select lang="de" value="option-1" clearable data-localization='{"clearEntry": "Reset!!"}'
```
karlbaumhauer, line 26:
```suggestion
      <sd-select lang="de" value="option-1" clearable data-custom-localization='{"clearEntry": "Reset!!"}'
```

karlbaumhauer: as agreed in the refinement.

mariohamann:
Currently all tests run locally:

CleanShot 2024-06-14 at 11 27 42@2x

mariohamann:
As Chromatic is currently not working, this is how the docs look like:

CleanShot 2024-06-14 at 11 33 40@2x

Vahid1919:
> As Chromatic is currently not working, this is how the docs look like. The important part here is "Custom Localization"
> 
 @mariohamann At the very beginning of the documentation, there seems to be a mistake regarding the example.
The code example seems to be rendering German translations, although the paragraph states it's an example for English.

packages/components/src/utilities/localize.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    `);
+    select.localize.setCustomLocalization({
+      numOptionsSelected: num => {
+        if (num === 0) return '';
+        return `Funds selected (${num})`;
+      }
+    });
+
+    // There's currently a bug that causes to not reactively change translated strings
+    // For this reason, we need to change the value first in this test
```
Vahid1919, line 97: is this bug related to the tests or coming from the component/localization? I am asking in case we need to create a ticket and add that to the comment.

Vahid1919 (approved):
Minor documentations errors. Besides that LGTM!

packages/components/src/docs/General/06-Localization.mdx
```diff
@@ -0,0 +1,168 @@
...
+    <script type="module" src="/path/to/es.js"></script>
+  </head>
+
+  <body>
+    ...
+  </body>
+</html>
+```
+
+:::tip
```
smfonseca, line 105: Small suggestion: you could emphasize the tip by displaying it as a blockquote

smfonseca (approved):

packages/components/src/utilities/localize.test.ts
mariohamann: There's already a ticket.

PR #966, opened 2024-04-19
feat: ✨ add sd-scrollable

azraefendic:
## Description:
Added the sd-scrollable component.

## Definition of Reviewable:
- [ ] Documentation is created/updated
- [ ] Migration Guide is created/updated
- [ ] E2E tests (features, a11y, bug fixes) are created/updated
- [ ] Stories (features, a11y) are created/updated
- [ ] relevant tickets are linked

PR #1007, opened 2024-05-02
feat: ✨ add sd-step & sd-step-group 

### LINKED ISSUE #304: feat: ✨ add sd-step & sd-step-group
## User Story
As a user of the Solid Design System, I want to use a sd-step-group component for smooth navigation through task steps.

## Design Branch v1.0.0
https://www.figma.com/file/VTztxQ5pWG7ARg8hCX6PfR/branch/XQ86g7rO9VAF1zCeb3fMlh/Solid-DS-%E2%80%93-Component-Library?type=design&node-id=19206-26790&mode=design&t=JudeeRyagcu3nT60-0

## Documentation
https://www.figma.com/file/RQSSKJKUI90k21fK8DDK3q/Progress-Steps?type=design&node-id=1202-9493&mode=design&t=fD5NNuUK6I7kSPez-0
## Hand-off
https://www.figma.com/file/RQSSKJKUI90k21fK8DDK3q/Progress-Steps?type=design&node-id=6-4&mode=design&t=fD5NNuUK6I7kSPez-0

#### Props
| <ul><li>[x] </li></ul>  | `size` | `lg` / `sm` | `lg` | Choose between `lg` for standard progress-steps or `sm` for tight spaces |
| <ul><li>[x] </li></ul>  | `orientation` | `horizontal` / `vertical` | `horizontal` | Choose between the 2 variants: horizontal & vertical. |
| <ul><li>[x] </li></ul>  | `state` | `waiting`/`finished`/`inProgress` | `waiting` | The indicator helps end users being aware of where they currently are within the process or task in relation to where they have been, and what to follow.  |
| <ul><li>[x] </li></ul>  | `noTail` | `boolean` | `false` | The border of the sd-step component.The border must extend 100% wide so that it can dock directly to the next sd-step component |
| <ul><li>[x] </li></ul>  | `label` | `string` | `` | Shows the label of the sd-step component. If you need to display HTML, use the `label` slot instead. |
| <ul><li>[x] </li></ul>  | `description` | `string` | `` | Shows the description of the sd-step component. If you need to display HTML, use the `description` slot instead. |

#### Parts (besides the shoelace parts)
  - [x] every other HTML element (depending on the HTML structure) => NOTE: if highlighting the parts in the `parts` story does not work out of the box, do not spend more effort on it but open a bug ticket to tackle it later.
#### Slots
  - [x] `label` - The progress-step label. Alternatively, you can use the `label` attribute.
  - [x] `default` - The sd-step description. Alternatively, you can use the `description` attribute.
  - [x] `Orientation` - Shows both orientation (horizontal & vertical)
  - [x] `Size x State` - Show in a table both sizes with the different states

#### Props
| <ul><li>[x] </li></ul>  | `size` | `lg` / `sm` | `lg` | The progress-steps group's size. This size will be applied to all child of sd-step |
| <ul><li>[x] </li></ul>  | `orientation` | `vertical` / `horizontal` | `horizontal` | The Group orientation. The orientation will be applied to all child of sd-step |

#### Parts (besides the shoelace parts)
  - [ ] every other HTML element (depending on the HTML structure) => NOTE: if highlighting the parts in the `parts` story does not work out of the box, do not spend more effort on it but open a bug ticket to tackle it later.
  - [x] `Size` - (description)
  - [x] `Orientation` - Shows both orientations (horizontal & vertical)
#### Samples
  - [ ] `name` - add sd-media with an image inside the default slot (check the sample section for Figma Design Branch v1.0.0 Link)

 #### Methods
| Status | Name | Description |
| ------------- | ------------- | ------------- |
| <ul><li>[x] </li></ul>  | `setActiveStep` | Sets the active step-item |

- [ ] Question1
- [ ] Question2

## Technical Information
- [x] Steps that are completed must be clickable. The state from the previous one should also change back to waiting (check the click bahaviour here: https://ant.design/components/steps#steps-demo-clickable)
- [ ] For the noTail property the pseudo-element ::after could be a thing
    - [x] decided to move away from pseudo-elements since they need absolute styles and those do not work well when steps are used together
- [x] Focus state needs to be done for the sd-step component (state: finished). Please refer to the Figma branch v1.0.0 link für further clarification.
- [x] you can use sd-paragraph for the label and default slot.

## Subtasks
- [x] #547
- [ ] #1031
- [ ] #1030

## DoR
- [x] Item has business value
- [x] All subtasks have been estimated by the team

## DoD

Vahid1919:
## Description:
Added sd-step and sd-step-group. Closes #304

## Definition of Reviewable:


packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+@customElement('sd-step')
+export default class SdStep extends SolidElement {
+  /** The step's size. */
+  @property({ reflect: true }) size: 'lg' | 'sm' = 'lg';
+
+  /** Determines the orientation of the step. */
+  @property({ reflect: true }) orientation: 'horizontal' | 'vertical' = 'horizontal';
+
+  /** Determines the state of the step. */
+  @property({ reflect: true }) state: 'waiting' | 'finished' | 'inProgress' = 'waiting';
```
yoezlem, line 40: Can we changed the order to => 'waiting' | 'inProgress' | 'finished' = 'waiting'; ?

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  /** Removes the tail from the step. */
+  @property({ reflect: true, type: Boolean, attribute: 'no-tail' }) noTail = false;
+
+  /** The step's label. */
```
yoezlem, line 45: Could you also add here that property label has a higher priority then the slot label?

yoezlem (commented):
image

1. The first sd-step has less margin (right side) than the last one (left side)
2. The left side of the inProgress sd-step (2) has more margin than the right side.

packages/components/src/docs/Migration/ui-progress-step.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+The new `[sd-step]` is designed to replace the `[ui-progress-step]`. Instead of mainly providing content via attributes, the `[sd-step]` component uses slots to allow for more flexibility and customization.
+
+## 💾 Slots
+
+### ✨ New Slots
+
+#### [default]
+
+The step's description.
```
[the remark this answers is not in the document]
karlbaumhauer: could also be added to the property/slot description but should be mentioned somewhere. Don't know how we did it with other components.

packages/components/src/docs/Migration/ui-progress-step.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+#### [label]
+
+The step's label.
```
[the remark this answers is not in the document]
karlbaumhauer: same here

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+  @property() label = '';
+
+  /** The step's description. */
```
karlbaumhauer, line 48:
```suggestion
  /** The step's description overwriting the `description` property and supposed to be used for complex description content. */
```

packages/components/src/components/step/step.ts
karlbaumhauer:
```suggestion
  /** The step's label overwriting the `label` property and supposed to be used for complex label content. */
```

karlbaumhauer: see commenrs in packages/components/src/docs/Migration/ui-progress-step.mdx

packages/components/src/components/step-group/step-group.stories.ts
```diff
@@ -0,0 +1,139 @@
...
+  render: (args: any) => {
+    return html`${generateTemplate({ args })}`;
+  }
+};
+
+/**
+ * This sample shows how to set the active step programmatically.
+ */
+
+export const SetActiveStep = {
```
karlbaumhauer, line 54: can you move this to the end? Samples are always the last stories in other components.

packages/components/src/docs/Migration/ui-progress-step.mdx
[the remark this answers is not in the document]
Vahid1919: We usually keep them similar. So I will go with that route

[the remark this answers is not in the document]
Vahid1919: There are minor variations between the main .ts file and the docs though, as the label attribute already exists in the ui-progress-step, it won't require additional docs in the migration guide. But, I will of course explain the nuances in the main file and storybook docs.

Vahid1919:
> When I am changing the tail from true to false and back to true, the tail didn't appear. Is there something wrong with Storybook or with the component?

@yoezlem I believe that's a Storybook issue, we have the same problem with the no-close-button attribute in sd-dialog. I remember having conversations with @mariohamann about this. But I cannot seem to find any explicit mentions on any issues or tickets.

van-nguyen-ht:
@Vahid1919 I left some comments :) just hand me back the ticket after you have a look ok?

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+              this.size === 'lg' ? 'w-12' : 'w-8',
+              this.state === 'waiting' && 'border-neutral-500',
+              this.state === 'finished' && 'border-primary hover:bg-primary-100 hover:border-primary-500',
+              this.state === 'inProgress' && 'bg-accent border-none text-white'
+            )}
+          >
+            ${
+              this.state === 'finished'
+                ? html` <sd-icon name="confirm" library="system" color="primary"></sd-icon>`
+                : this.StepIndex
```
mariohamann, line 114: Please make this a slot as well. People should be able to override this icons or even set their own content or whatever. Maybe you'll need two slots, `<slot name="complete-icon">`, `<slot name="index">`

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+  @property({ reflect: true, type: Boolean, attribute: 'no-tail' }) noTail = false;
+
+  /** The step's label overwriting the `label` slot. Use the `label` slot for complex label content. */
+  @property() label = '';
+
+  /** The step's description overwriting the `description` slot. Use the `description` slot for complex description content. */
+  @property() description = '';
+
+  /** The step's number in a step-group */
+  @property({ type: Number, reflect: true, attribute: 'step-index' }) StepIndex = 1;
```
mariohamann, line 52: Call it only index.

packages/components/src/components/step-group/step-group.ts
```diff
@@ -0,0 +1,117 @@
...
+        step.StepIndex = index + 1;
+        step.size = this.size;
+        step.orientation = this.orientation;
+
+        // Set the active step on click
+        step.addEventListener('click', () => {
+          if (step.state === 'finished') {
+            this.setActiveStep(index);
+          }
+        });
```
mariohamann, line 57: This should definitely be removed as it is way to invasive. We should never handle the interaction here, as there will usually state management that is in charge of that. It could even happen, that there is data, that is not correct etc. or they shouldn't be allowed to go backt or whatever.

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+              ? this.size === 'lg'
+                ? 'translateLg'
+                : 'translateSm'
+              : this.size === 'lg'
+                ? 'mt-1'
+                : 'mt-3'
+          )}
+        >
+          
+          <${tag}
```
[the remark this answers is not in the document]
mariohamann: See e. g. https://codepen.io/pen?&editors=001 where everyrthing is clickable.

packages/components/src/components/step-group/step-group.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  private getAllSteps() {
+    return [...this.body.assignedElements()].filter(el => el.tagName.toLowerCase() === 'sd-step') as [SdStep];
+  }
+
+  /**
+   * Sets the active step.
+   * @param index The index of the step to set as active.
+   */
+  setActiveStep(index: number) {
```
mariohamann, line 84: This could be a public method, right?

packages/components/src/components/step/step.ts
mariohamann: See a real life example from the website where the index should be something different:

CleanShot 2024-05-03 at 16 26 31@2x

mariohamann: CleanShot 2024-05-03 at 16 29 27@2x

packages/components/src/components/step-group/step-group.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  render() {
+    return html`
+      <div part="base" class=${cx('flex', this.orientation === 'vertical' && 'flex-col h-full')}>
+        <slot part="body"></slot>
+      </div>
+    `;
+  }
```
mariohamann, line 100: We maybe should show the component with `sd-scrollable`, see here: https://cms-components.dev.fe.union-investment.de/integration/storybook/index.html?path=/story/cms-components-cms-progress-bar--configurable&args=activeStepNumber:8&globals=backgrounds.value:!hex(F6F6F6)

packages/components/src/components/step-group/step-group.stories.ts
```diff
@@ -0,0 +1,139 @@
...
+          Lorem ipsum est dolor sit amet
+        </sd-step>
+
+        <sd-step size="lg" orientation="horizontal" state="waiting">
+          <p slot="label">Reprehenderit qui in e name</p>
+          Lorem ipsum est dolor sit amet
+        </sd-step>
+      </sd-step-group>
+
+      <sd-button class="w-min mt-8" size="sm" id="next">Next</sd-button>
```
mariohamann, line 125: Please make a `back` button, too. The UX isn't fun that you have to up to go to a step back.

mariohamann (changes requested):

packages/components/src/components/step/step.ts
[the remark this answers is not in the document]
mariohamann: In CMS context sometimes even everything is NOT clickable and just informational. THis should be possible, too.

mariohamann:
Please re-assign when done.

packages/components/src/components/step-group/step-group.stories.ts
karlbaumhauer: Would not agree.
The idea here was to show that people can click on the completed step to directly jump back to it. A back button would make the whole story obsolete IMO.

maybe we can refactor the description here to make clear what we want to show?!

packages/components/src/components/step/step.ts
```diff
@@ -0,0 +1,176 @@
...
+@customElement('sd-step')
+export default class SdStep extends SolidElement {
+  /** The step's size. */
+  @property({ reflect: true }) size: 'lg' | 'sm' = 'lg';
+
+  /** Determines the orientation of the step. */
+  @property({ reflect: true }) orientation: 'horizontal' | 'vertical' = 'horizontal';
+
+  /** Determines the state of the step. */
+  @property({ reflect: true }) state: 'waiting' | 'inProgress' | 'finished' = 'waiting';
```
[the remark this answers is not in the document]
karlbaumhauer: I don't see the need to change it to something less component specific as the component will only be used for this context. If we want to change it to something with less context (which I would be totally fine with), we need to change it in design too and I would then go further to something like: 
- past
- current
- upcoming

karlbaumhauer (approved):
besides the comments from @mariohamann, my original change requests look good. thx.

packages/components/src/components/step-group/step-group.ts
Vahid1919: yes. It is public. Usage is shown in the sample.

packages/components/src/components/step-group/step-group.ts
Vahid1919: @mariohamann We have the requirement that "finished" steps should be clickable. 
from #304 ->
 "Steps that are completed must be clickable. The state from the previous one should also change back to waiting (check the click bahaviour here: https://ant.design/components/steps#steps-demo-clickable)"
 
 The ant.design demo moves to the clicked step. Are you proposing we remove the functionality as a whole?

packages/components/src/components/step/step.ts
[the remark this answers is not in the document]
Vahid1919: I personally don't mind the states being component specific, if they describe the role more accurately. I hesitate to use 'disabled' for example as that provides no information about the positioning of the step, while 'current' and 'complete' imply an order. 

I would prefer the original state names we used. Thoughts? @mariohamann @karlbaumhauer

packages/components/src/components/step-group/step-group.stories.ts
Vahid1919: I think it could be useful to include the back button since the story is mainly to show that you can programatically change the steps. 

Instead I could add the information about clicking to the description (I think a description alone could be enough as we now have hover styles on finished steps that indicate "click-ability" anyways)

karlbaumhauer: in general, if we talk about UX, we could include our designers into this... ;-)

In specific, I like the idea to only have the description. It makes it even more clear, that we don't talk about a functionality we provide but one we want to show...

packages/components/src/components/step/step.ts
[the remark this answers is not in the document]
karlbaumhauer: @mariohamann what would be the reason to have it non-component-specific? Do you see the component being used for something else but classic steps? 

Mayebe we should clarify this first before we try to find a wording which sounds better... 🤣 @coraliefeil / @EllaELSA any thoughts on this?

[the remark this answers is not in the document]
mariohamann: Thanks for all the thoughts, that helped me to have further thoughts on that. Let's take a step back:

Step groups in general can be `interactive` and `non-interactive` – if they are `non-interactive`, they usually all have the same state, which is `blue`.

For the discussion here we should focus on `interactive` elements. Here the `step-group` serves as a navigation, having `role=navigation` (if there are interactive elements – @Vahid1919!). Inside that, we either have `links` or `buttons`. There are roles to influence the semantics of these items, which we use in 'sd-navigation-item`: `disabled` and `current`. The latter e. g. provides a specific attribute which in this case should be `aria-current="step"` (https://www.accede-web.com/en/guidelines/html-css/complementary-guidelines/identify-the-current-step-in-multiple-steps-forms-using-aria-currentstep/).

Let's make a comparison here:

CleanShot 2024-05-07 at 16 29 42

Seeing this, taking a11y, semantics and consistency into account, and after our talk @Vahid1919, I therefore would propose:

- `disabled` → boolean to make it look disabled and make it semantically a disabled button
- `current` → boolean to make it behave look and from aria-roles `current`

If nothing is set, it's just behaving as "default".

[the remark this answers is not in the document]
mariohamann: I think this makes especially sense if applications decide, that all elements in the steps should be clickable – which could totally be a case! Then everything besides `current` has the default state and there even is now `disabled` state. I don't see a reason, why this should be prevented @EllaELSA @coraliefeil? It could be problematic for forms, but if you think about interactive mini-apps like "Get to know more about process to buy funds by clicking on the steps" ore similar. (I hope this example is some understandable. 😆 )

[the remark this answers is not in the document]
karlbaumhauer: thx @mariohamann for the explanation. looking at it from a navigation poit of view and not from a steps perspective, I agree to the suggestion of default, current and disabled. 

I would just love to clarify your last question towards @EllaELSA (in terms of form elements) and in any case add this information to the documentation. Just from the naming og the component it sounds strange but with this information, it is more clear and also shows the generic structure of our components.

packages/components/src/components/step-group/step-group.ts
Vahid1919: Sounds good. But I will have to wait for #966 to be merged.

packages/components/src/components/icon/library.system.ts
```diff
@@ -60,7 +60,7 @@ export const icons = {
   `,
   'info-circle': `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M12 2C6.477 2 2 6.477 2 12s4.477 10 10 10 10-4.477 10-10S17.523 2 12 2Zm0 18a8 8 0 1 1 0-16 8 8 0 0 1 0 16Z"/><path d="M13.333 7.833a1.5 1.5 0 1 1-3.001-.001 1.5 1.5 0 0 1 3.001.001ZM13 11.667a1 1 0 0 0-1-1h-1a1 1 0 0 0 0 2v2.667h2v-3.667Z"/><path d="M11 16.333v-1h-1a1 1 0 0 0 0 2h2a1 1 0 0 1-1-1Zm3-1h-1v1a1 1 0 0 1-1 1h2a1 1 0 0 0 0-2Z"/><path  d="M13 16.333v-1h-2v1a1 1 0 0 0 2 0Z"/></svg>`,
   'status-hook': `
-    <svg xmlns="http://www.w3.org/2000/svg" width="12" height="12" fill="none">
+    <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 12 13">
```
mariohamann, line 63: Is 0 0 12 13 correct?

packages/components/src/components/step/step.ts
```diff
@@ -37,19 +41,22 @@ export default class SdStep extends SolidElement {
   @property({ reflect: true }) orientation: 'horizontal' | 'vertical' = 'horizontal';
 
   /** Determines the state of the step. */
-  @property({ reflect: true }) state: 'waiting' | 'inProgress' | 'finished' = 'waiting';
+  @property({ reflect: true }) state: 'disabled' | 'current' | 'default' = 'default';
```
mariohamann, line 44: Looks great in general – just one note here coming from the investigation: https://github.com/solid-design-system/solid/pull/1007#discussion_r1592590947

Instead of "state" with options, introduce `current` and `disabled` as booleans, so it fits to our other components.

mariohamann (changes requested):

packages/components/src/components/icon/library.system.ts
Vahid1919: @mariohamann I set those values manually since the checkmark path is not centered. it's 12 13 since the width is larger. I made sure with Van that this change leaves the other components that use the icon unchanged.

mariohamann: Fine!

coraliefeil:
Good job! Pls reassign for another review.

packages/components/src/components/icon/library.system.ts
```diff
@@ -40,11 +40,6 @@ export const icons = {
       <path d="M12 2C6.477 2 2 6.477 2 12s4.477 10 10 10 10-4.477 10-10S17.523 2 12 2Zm0 18a8 8 0 1 1 0-16 8 8 0 0 1 0 16Z"/>
       <path d="M16.373 7.627a1 1 0 0 0-1.414 0l-2.96 2.96-2.96-2.96a1 1 0 0 0-1.413 1.414l-.001-.001 2.96 2.96-2.96 2.96a1 1 0 0 0 1.414 1.414l2.96-2.96 2.96 2.96a1 1 0 0 0 1.414-1.414L13.413 12l2.96-2.96a1 1 0 0 0 0-1.414v.001Z"/>
     </svg>`,
-  confirm: `
-    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
-      <path d="M21.528 2.167a1 1 0 0 0-1.384.277l-.002.004-11.333 17-5.127-5.153a1 1 0 0 0-1.413 1.414l-.001-.001 6 6a.997.997 0 0 0 .707.293h.1c.307-.032.57-.199.731-.44l.002-.004 12-18a1 1 0 0 0-.276-1.388l-.004-.002Z"/>
-    </svg>
-  `,
```
Vahid1919, line 47: @mariohamann @karlbaumhauer 
I wanted to remove this confirm icon as it is redundant (status-check already exists, and is preferred by design). 

Would this constitute as a breaking change?

karlbaumhauer: thats the question. No one should but everyone could and therewith it IMO is a breaking change by definition. Not by whish... ;-)

karlbaumhauer: but I am not 100% sure...

Vahid1919: I took a look at our documentation that also indicates that system icons should not be used by themselves. 
image

If we go the breaking change route,by effect, we would be supporting a use case we don't recommend. As this is only a breaking change if you are using the icons where they are not supposed to be. 

I would lean on the side of **not making it a breaking change** and keep an eye out for any issues. Ideally, no one is using them and we can keep treating it as strictly internal.

 Otherwise, we can communicate the change to the respective users that bring this up as an issue, and recommend they replace the icons or we can reevaluate the process then. That's my suggestion. What do you think? @karlbaumhauer @mariohamann

karlbaumhauer: thx @Vahid1919. I second what you wrote but still are not 100% sure here... As you described it correctly, it would only be a breaking change if people use our library in a wrong way. 

I asked copilot and got the same answer...🤣
Screenshot 2024-05-22 at 13 58 48

The only thing which hinders my accepting it as non-breaking:
- we only "recommend" not to use it (instead of not "allowing" it). IMO we, at least, should write that they should not be used and will not be considered by semantic release.
- our changes are not backwards compatible and therewith by definition a breaking change.

Don't get me wrong. Content wise, I totally agree to what you both wrote. I am just struggling with the fact that we haven't been very clear in our docs on that, to just say: "not our issue".

mariohamann: I think it's very clear: "As these may change over time, we don't recommend using them directly." I'm clearly on the side on not making it a breaking change.

Otherwise everything could be a breaking change, e. g. private methods which are renamed (but which can still be used), changed markup etc. This would prevent us from making changes at all.

From my opinion it's about how things are documented: CSS Variables, Public Methods etc. are clearly documented to be used – if we change those, it's breaking. Everything else happens under the hood, and that's fine.

karlbaumhauer: I am fine with that (and like it) in comparison to private methods.

the only request I would have (for future changes): lets reformulate the docs on that topic. I do not agree, that we are "very clear" as @mariohamann stated. I would love to see something like: "you should never use it as we don't take internal icons into account when it comes to backwards compatibility"...

mariohamann (approved):

coraliefeil:
Really nice ;)
Pls reassign for final review

coraliefeil:
Cool stuff. Approved. ;)

packages/components/src/components/step-group/step-group.ts
Vahid1919: I will create a follow-up ticket for this :)

Issue #867, opened 2024-03-14
fix: 🤔 replace video with placeholder video


coraliefeil:
## Current behavior

1. The [provided video](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/story/components-sd-video--video-element) is not the one from the placeholder package.

2. I cannot turn on subtitles -> How does it work? How do subtitles look like? 

## Expected behavior
Show this video from the placeholder package: (#378 )
https://github.com/solid-design-system/solid/issues/867#issuecomment-2058947661

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [x] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [x] Relevant stories (Features, A11y) are created/updated
- [x] Implementation works successfully on `feature` branch

yoezlem:
@EllaELSA could you put a video with audio track in it?

EllaELSA:
@coraliefeil I had a chat with Brand regarding this. She suggested the Corporate Design explanatory film from the toolbox. Does this work for you? let me know then I can try to get it 
[CD-Toolbox (union-investment.de)](https://cd.union-investment.de/)

EllaELSA:
I have the mp4 now 

https://github.com/solid-design-system/solid/assets/124053512/8265bd18-fd4e-4edc-bd54-1c3504f38080

yoezlem:
@EllaELSA Could you send the audio again? Somethings not right with it.

yoezlem:
@EllaELSA nvm its working now. Maybe I had slow Internet connection yesterday.

yoezlem:
@coraliefeil I added the video in our placeholder package. For subtitles we need the `<track>` attribute and a .vtt file: https://developer.mozilla.org/en-US/docs/Web/Media/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video#html_markup

We can add this to a new task and I would close this one after https://github.com/solid-design-system/solid/pull/1032 is reviewed

yoezlem:
Looks like a duplicate to -> [#378](https://github.com/solid-design-system/solid/issues/378)

yoezlem:
For subtitles and subtitles menu we need styles I believe. 

[...]

Issue #378, opened 2023-09-13
feat: ✨ extend placeholders package with videos


christophsaile:
## User Story
As a User of the Solid Design System, I would like to have videos in the placeholder package,
so that I don't have to rely on other dummy data.

### Suggesed Solution
Add a video to the placeholder package. https://github.com/solid-design-system/solid/issues/867#issuecomment-2058947661

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

coraliefeil:
@EllaELSA if we receive a placeholder video from Brand, pls forward to @karlbaumhauer for implementation. Thanks a lot.

christophsaile:
@EllaELSA any updates on this?

EllaELSA:
@christophsaile please use the video Debbi got for the infinity comp. I got an approval for this from Sandra. Just waiting for the url of the video from Eileen.
Will update this on monday

EllaELSA:
@christophsaile the video we can use ist this one.
https://www.pexels.com/de-de/video/liebe-menschen-garten-zeitlupe-4625767/

karlbaumhauer:
@EllaELSA I had to delete a your comment as you mentioned personal information from people outside this team. Pls make sure you do not provide any personal information in this public repository.

coraliefeil:
@EllaELSA can we offer a video with audio track in addition?

karlbaumhauer:
@EllaELSA What is the current status here? What would be the size of the ticket?

EllaELSA:
Brand suggested the explanatory film from the CD toolbox. If this works for you I can try to get the file. @karlbaumhauer

karlbaumhauer:
This would be benefitial for a start, IMO.

yoezlem:
closed with #1032

PR #1032, opened 2024-05-03
feat: ✨add placeholder video

yoezlem:
## Description:
#867 & #378 

## Definition of Reviewable:

### DIFF
--- a/packages/components/.storybook/main.cjs
+++ b/packages/components/.storybook/main.cjs
@@ -19,7 +19,11 @@ module.exports = {
     './assets',
     {
       from: '../../placeholders/src/images',
-      to: '/placeholders'
+      to: '/placeholders/images'
+    },
+    {
+      from: '../../placeholders/src/videos',
+      to: '/placeholders/videos'
     },
     {
       from: '../node_modules/countup.js/dist',
--- a/packages/components/src/components/teaser-media/teaser-media.stories.ts
+++ b/packages/components/src/components/teaser-media/teaser-media.stories.ts
@@ -63,7 +63,7 @@ export const Variant = {
           type: 'slot',
           name: 'media',
           value:
-            '<img slot="media" class="aspect-video object-cover" src="./placeholders/generic.jpg" alt="A skyline of a city by night"/>'
+            '<img slot="media" class="aspect-video object-cover" src="./placeholders/images/generic.jpg" alt="A skyline of a city by night"/>'
         }
       ]
     });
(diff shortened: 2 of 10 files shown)


karlbaumhauer (approved):

mariohamann (changes requested):
This is not a breaking change! Even not for the placeholder package.

packages/components/.storybook/main.cjs
```diff
@@ -19,7 +19,11 @@ module.exports = {
     './assets',
     {
       from: '../../placeholders/src/images',
-      to: '/placeholders'
+      to: '/placeholders/images'
+    },
+    {
+      from: '../../placeholders/src/videos',
+      to: '/placeholders/videos'
```
mariohamann, line 26: This is not a breaking change, as the original paths are unchanged.

mariohamann (changes requested):

packages/components/.storybook/main.cjs
karlbaumhauer: good point but looking at the changes below, users would need to change the path as we had to in the stories... Or what to we both miss... ;-)

mariohamann: ?

"from" didn't change and there was just added a new video. This feature has nothing to do with Storybook or how others configure it.

There is 0 change for people using the placeholder package.

yoezlem: yeah, the from is the same but the to path is now different for images. Thats why I though it might be a Breaking Change. 
So if someone is using our placeholder images and we changing the to path.. would that not break their application?

mariohamann: The package didn't change the path. It's still `/images`. We added in OUR components docs `/videos`, but that has nothing to do how others use the package.

mariohamann: In that case it's especially harmful, as Semantic Release would even increase do a major release for the components package. (A readon, why we should think about separating docs from components in future AND maybe using `changesets`.)

yoezlem: Allright, I removed the description.

mariohamann (approved):

packages/components/.storybook/main.cjs
karlbaumhauer: my bad... mixed up the files and overlooked, that we are talking about storybook config. Sorry for that.

Issue #719, opened 2024-01-29
fix: 🤔 sd-textarea doesn't accept full height of container


christophsaile:
## Current behavior
if you place a sd-textarea with height:100% in a container the sd-textarea won't take the height of the container.

## Expected behavior
I would expect that the input field of the sd-textarea has the same height as the container.

## Steps to reproduce
https://jsfiddle.net/f3krvean/1/

## Technical Information
without part-selectors:
![grafik](https://github.com/solid-design-system/solid/assets/54807480/c0e4c75a-fb54-47f9-bb26-8ba397606610)

fix with part-selectors:
![grafik](https://github.com/solid-design-system/solid/assets/54807480/4096fadb-dc01-4a2f-9650-db1c7faecd58)
![grafik](https://github.com/solid-design-system/solid/assets/54807480/22b2f18d-1907-49e4-a024-74573d824b18)

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

karlbaumhauer:
@christophsaile as this is rather a special use-case and the component currently aligns with text rows, we are tending not to implement it into the component.

Do you have more insights where this feature could be useful? If you like, you could provide a sample in a feature branch and we can add it to the samples section. therewith users could easily copy-paste this feature/use-case.

christophsaile:
@karlbaumhauer in my opinion this feature is always usefull if you want to align a text area next to another element. As you can see in my screenshot it looks wired if the text area does not have the same height as the element next to it. Of course you can add more rows so it gets closer to the size of the left element but still you would not match 100% the height.
+ Nativ text-area input also takes 100% of container.

karlbaumhauer:
@christophsaile we decided to add this to the component. As this is not part of the CL-migration, it will come in Q2 unless you send us a PR... ;-)

yoezlem:
I will put that again in ready and will close the PR. Branch is open, if someone has the time to work on that, feel free!

PR #1036, opened 2024-05-03
fix: accept full height of container

yoezlem:
## Description:
Changed the logic regarding #719 to accept full height inside a container 

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/textarea/textarea.ts
+++ b/packages/components/src/components/textarea/textarea.ts
@@ -333,7 +333,7 @@ export default class SdTextarea extends SolidElement implements SolidFormControl
     }[this.size];
 
     return html`
-      <div part="form-control" class="text-left">
+      <div part="form-control" class="flex flex-col w-full text-left">
         <label
           part="form-control-label"
           id="label"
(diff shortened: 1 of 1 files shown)


packages/components/src/components/textarea/textarea.ts
```diff
@@ -333,7 +333,7 @@ export default class SdTextarea extends SolidElement implements SolidFormControl
     }[this.size];
 
     return html`
-      <div part="form-control" class="text-left">
+      <div part="form-control" class="flex flex-col w-full text-left">
```
mariohamann, line 336: Does this fully preserve the ability to set rows? When does it use rows, when the height? Some documentation or even better a story would be very helpful here.

mariohamann (changes requested):

mariohamann:
I removed myself – feel free to re-add if needed.

packages/components/src/components/textarea/textarea.ts
mariohamann: @yoezlem Anything you need here? :)

Vahid1919 (approved):

azraefendic (approved):

PR #1121, opened 2024-05-30
feat: ✨ add sd-flipcard

### LINKED ISSUE #1012: feat[dev]: ✨ implement sd-flipcard component
## User Story
As a user of the Solid Design System, I want to utilize a Flashcard component with flip animation to add dynamic and interactive content to pages, providing a visually engaging way to present and interact with information.

## Documentation
https://www.figma.com/file/JZZL5QPXQVKVGt4CHBlNlz/Flashcard?type=design&node-id=1301-7360&mode=design&t=eaEgzwqpgma861CK-0

## Hand-off
https://www.figma.com/file/JZZL5QPXQVKVGt4CHBlNlz/Flashcard?type=design&node-id=6-4&mode=design&t=eaEgzwqpgma861CK-0

#### Props
| <ul><li>[ ] </li></ul>  | `frontVariant` | `primary` / `primary-100` / `gradient light-top` / `gradient light-bottom` / `gradient dark-top` / `gradient dark-bottom`| `primary` | The sd-flashcards theme variant |
| <ul><li>[ ] </li></ul>  | `backVariant` | `primary` / `primary-100` / `gradient light-top` / `gradient light-bottom` / `gradient dark-top` / `gradient dark-bottom`| `primary` | The sd-flashcards theme variant |
| <ul><li>[ ] </li></ul>  | `ratio` | `3:4` / `16:9` | `3:4` | the ratio for the layout of the flashcard |

#### CSS-Properties
  - [ ] `name` - description
#### Parts (besides the shoelace parts)
  - [ ] every other HTML element (depending on the HTML structure) => NOTE: if highlighting the parts in the `parts` story does not work out of the box, do not spend more effort on it but open a bug ticket to tackle it later.
#### Slots
  - [ ] `front` - for the content on the front side of the flashcard
  - [ ] `back` - for the content of the back side of the flashcard
  - [ ] `mediaFront` - An optional media slot which can be as a background. Dependent from gradient variant.
  - [ ] `mediaBack` - An optional media slot which can be as a background. Dependent from gradient variant.
  - [ ] `Variant` - (description)
  - [ ] `Ratio` - (description)
#### Samples
  - [ ] `Clickable Flashcard` - Only it is necessary that is allowed to do it semantically (accessible way).

 #### Methods
| Status | Name | Description |
| ------------- | ------------- | ------------- |
| <ul><li>[ ] </li></ul>  | `flip-front` | Shows the content of the quickfact. |
| <ul><li>[ ] </li></ul>  | `flip-back` | Hides the content of the quickfact. |

 #### Events
| Status | Name | Description |
| ------------- | ------------- | ------------- |
| <ul><li>[ ] </li></ul>  | `sd-flip-front` | Shows the content of the quickfact. |
| <ul><li>[ ] </li></ul>  | `sd-flip-back` | Hides the content of the quickfact. |

- [ ] Question1
- [ ] Question2

### Technical Information
- [ ] transition from back/flip is 1000ms
- [x] We have to check if its allowed to add a link inside a button (accessiblity perspective). Generally we have to check nested elements. Please document that. 
- [x] we have to make sure that both sides are searchable and available in the DOM. Maybe we can use the `summary` and `details` from accordion.

## DoR
- [x] Item has business value
- [x] All subtasks have been estimated by the team

## DoD
- [x] All subtasks have been closed

## DoR

## DoD

Vahid1919:
## Description:

Added sd-flipcard implementation .Closes  #1012. 

## Definition of Reviewable:


packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,277 @@
...
+    | 'gradient-dark-bottom' = 'primary-100';
+
+  /** Determines the ratio for the layout of the flipcard. */
+  @property({ type: String, reflect: true }) ratio: '3:4' | '16:9' = '3:4';
+
+  connectedCallback() {
+    super.connectedCallback();
+  }
+
+  private flipFront() {
```
karlbaumhauer, line 65: From my perspective, the focus should switch to the back at the same time this function has been called. Currently the focus stays on the front element.

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  private flipBack() {
+    this.front.classList.remove('clicked--front');
+    this.back.classList.remove('clicked--back');
+    this.emit('sd-flip-back');
+  }
+
+  private handleFrontKeydown(event: KeyboardEvent) {
+    if (event.code === 'Enter' || event.code === 'Space') {
```
karlbaumhauer, line 78: I am not sure if this is the same case on all devices but on my mac, using MS Edge browser, the space bar also triggers a scroll down. This leads to the issue the flipcard moving out of my screen and flipped... 

Not sure if this needs to be a general issue in a new ticket as it might be the case in every keyboard controlled component and I am not sure about the WCAG requirements with the space bar, but I recognized it here again.

Enter key works fine.

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,277 @@
...
+    super.connectedCallback();
+  }
+
+  private flipFront() {
+    this.front.classList.add('clicked--front');
+    this.back.classList.add('clicked--back');
+    this.emit('sd-flip-front');
+  }
+
+  private flipBack() {
```
karlbaumhauer, line 71: same here. Focus should be sitched back to the front as well.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+  render: (args: any) => {
+    return html`<div style="width:300px">${generateTemplate({ args })}</div>`;
+  }
+};
+
+/**
+ * The sd-flipcard can be displayed in several ways using the `frontVariant` and `backVariant` attributes. This example shows the usage `frontVariant` attribute.
+ */
+
+export const Variants = {
```
karlbaumhauer, line 57: there seem to be a bug in the primary variant, as I get it displayed 2 times 
image

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+        }
+      ]
+    })
+};
+
+/**
+ * The sd-flipcard can be displayed in 3:4 or 16:9 aspect ratio.
+ */
+
+export const Ratio = {
```
karlbaumhauer, line 82: while switching the sides, my browser adss a scrollbar and therewith I have some layout shifting. 
Only happens when I switch the 3:4 variant in this story.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+          }
+        </style>
+        ${story()}`
+  ] as unknown
+};
+
+/**
+ * Use the `front`, `back`, `mediaFront` and `mediaBack` slots to add content to the flipcard.
+ */
+export const Slots = {
```
karlbaumhauer, line 108: for front and back, it shows a strange size of the slot. Probably correct but looks buggy.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+        });
+      })}
+    `;
+  }
+};
+
+/**
+ * Use the `base`, `front`, `back`, `frontSlotContainer`, `backSlotContainer`, `mediaFront`, `mediaBack`, `frontSecondaryGradient` and `backSecondaryGradient` parts to style the flipcard.
+ */
+export const Parts = {
```
karlbaumhauer, line 156: the `mediaBack` part is not visible as the sample only shows an image on the front.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+        },
+        {
+          type: 'attribute',
+          name: 'frontVariant',
+          value: 'gradient-dark-top'
+        },
+        {
+          type: 'attribute',
+          name: 'darkVariant',
+          value: 'gradient-dark-bottom'
```
karlbaumhauer, line 211: `backSecondaryGradient` is also not visible as the back only has `primary100` without any media or gradient.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+      ]
+    });
+  }
+};
+
+/**
+ * `sd-flipcard` is fully accessibile via keyboard.
+ */
+
+export const Mouseless = {
```
karlbaumhauer, line 222: is it possible to prevent imidiate switch, when opening the story? Its most probably as we change the focus on story-load but this feels buggy as well.

packages/components/src/components/flipcard/flipcard.stories.ts
```diff
@@ -0,0 +1,275 @@
...
+      pointerEventsCheck: 0
+    });
+  }
+};
+
+/**
+ * Here is a sample of the `sd-flipcard` with custom content in the `front` and `back` slots.
+ */
+
+export const Sample = {
```
karlbaumhauer, line 243: this is still WIP right? If not, we need to discuss the UX here.
As the flipcard flips on hover, you never will be able to click the link on the front... This does not make sense at all.

@coraliefeil was the `flip-on-hover` actually a result from the refinement? I see this as a major issue especially as I have in mind, automatic behavior needs to be stopped if necessary (WCAG requirement?!?). Users need to decied on wich speed they want the content to change (not the duration but the point where it should start, like autoplay stuff with media).

packages/components/src/components/flipcard/flipcard.test.ts
karlbaumhauer, line 1: pls add some tests.

karlbaumhauer (changes requested):

packages/components/src/components/flipcard/flipcard.ts
mariohamann: I'm especially worried: If there are interactive elements – how should this work? So to make this accessible via keyboard, focus should stay on the whole flip card to be able to flip back immediately, right? But actually when the whole thing is tappable, I'm not sure if we can add nested things to tab as well?

Or should we maybe introduce a tabbable button which only shows up on focus?

This will be a tough one...

packages/components/src/components/flipcard/flipcard.stories.ts
mariohamann: As this is user controlled, I believe it's fine. But we should definitely consider to turn off the animation here or make a simpler one instead, like a simple dissolve for people with reduced motion.

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,277 @@
...
+  @property({ type: String, reflect: true }) backVariant:
+    | 'primary'
+    | 'primary-100'
+    | 'gradient-light-top'
+    | 'gradient-light-bottom'
+    | 'gradient-dark-top'
+    | 'gradient-dark-bottom' = 'primary-100';
+
+  /** Determines the ratio for the layout of the flipcard. */
+  @property({ type: String, reflect: true }) ratio: '3:4' | '16:9' = '3:4';
```
mariohamann, line 59: Let's please remove this.

This is easily doable with one line of CSS. I don't see a reason why we should set this via JS. https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,277 @@
...
+    | 'gradient-dark-bottom' = 'primary';
+
+  /** Determines the variant of the back face of the flipcard. */
+  @property({ type: String, reflect: true }) backVariant:
+    | 'primary'
+    | 'primary-100'
+    | 'gradient-light-top'
+    | 'gradient-light-bottom'
+    | 'gradient-dark-top'
+    | 'gradient-dark-bottom' = 'primary-100';
```
mariohamann, line 56: TBH this whole thing worries me a lot.

For my feeling this component gives too many stylistic opinions. They might make sense in CMS context. I'm not sure if we should instead give these options more to the container or make a gradient container or similar which could then be slotted.

Dunno. Not really happy with how this evolved.

mariohamann: I just realized, that I raised my perspective on that: https://github.com/solid-design-system/solid/issues/1012#issuecomment-2144579151

Not sure, why this was not taken into consideration. Let's please align on that @yoezlem but we will do here quite some double work.

mariohamann:
@Vahid1919 @coraliefeil @yoezlem  So here's my proposal:

1. Let's remove the aspect-ratio, but instead set `aspect-ratio: 3 / 4` on `:host`, so we have a nice default. From that on let's show a story, that it's overridable with e. g. `<sd-flipcard style="aspect-ratio: 16 / 9">`.
2. Let's introduce `empty` for `front` and `back` as default – which has no padding and anything. This makes it possible to use the flipcard with whatever needed. We leave the other options as they are, so that CMS can easily just pick them as they need.

packages/components/src/components/flipcard/flipcard.ts
Vahid1919: I am okay with removing the flip behaviour on Space and only limiting to the "Enter" key. I thought of the spacebar as more of a secondary button anyways.

karlbaumhauer: we realized yesterday, that it seems to be a user setting on my laptop, so I wouldn't remove it directly but research which keys should be / need to be activated for accessibility.

packages/components/src/components/flipcard/flipcard.stories.ts
Vahid1919: Definitely a WIP @karlbaumhauer . This story did need some discussion. 

Mainly, I think it is about setting the flip-on-hover behaviours as a default. As this is clearly problematic anywhere where a user interaction is needed. I don't think any thing should be triggered by a hover when there are links/buttons present. 

I am not sure how the simpler animation would help here as users would still have issues interacting with the first page. (Maybe I am missing something here @mariohamann ) I see two possibilities here: 

Option 1: We remove the flip-on-hover behaviour and introduce an attribute to add this behaviour should users want it. So by default only clicking would work. 

Option 2: If we want to keep the flip-on-hover behaviour, we can add something like an 'interactive' attribute that disables flipping for hover.

I am leaning towards option 2 now. But what do you think @karlbaumhauer @mariohamann @coraliefeil ?

packages/components/src/components/flipcard/flipcard.stories.ts
Vahid1919: removed the ratio story in it's entirety due to refactoring. @karlbaumhauer please let me know if this happens in the other stories.

packages/components/src/components/flipcard/flipcard.ts
Vahid1919: I added the focus feedback on the latest iteration. Please take a look, especially on the sample to see how the flip-card behaves when there is tabbable content inside of the cards. @mariohamann @karlbaumhauer

packages/components/src/components/flipcard/flipcard.stories.ts
Vahid1919: This is an emergent bug created with the interaction of our formatter and Storybook. Since the number of options is large prettier formats it like this: 
image


Storybook interprets  the empty space before the first pipe(|) as an option, hence resulting in this duplication. This is solely a Storybook issue. I currently can't find a way to skip formatting for a single line, so I would suggest we leave this as a WON'T FIX. What do you think? @karlbaumhauer @mariohamann

packages/components/src/components/flipcard/flipcard.stories.ts
karlbaumhauer: What was the solution here as it is marked as "resolved"? I still think the hover behavior is quite a pain as we still have to make it by click for mobile/touch devices.

packages/components/src/components/flipcard/flipcard.stories.ts
karlbaumhauer: any solution idea @mariohamann?
@Vahid1919 if you cant fix it, pls open a ticket and put it into the backlog.

packages/components/src/components/flipcard/flipcard.stories.ts
karlbaumhauer: forget it, saw the new comment above.

packages/components/src/components/flipcard/flipcard.ts
karlbaumhauer: it works from my side even though I am still not happy with the hover... ;-) nevertheless, this should be decided by UX but for now I am fine.

karlbaumhauer (approved):
@Vahid1919: approved from my side.

disclaimer: I am currently not able to give some more detailed feedback. The points I mentioned in my last review seem all to be resolved. If you still need some feedback on specific topics, pls let me know.

packages/components/src/components/flipcard/flipcard.ts
karlbaumhauer: the flip-on-hover in general. But it might be a very personal oppinion.

smfonseca (commented):
@Vahid1919 I tested the component locally and specifically in the `Sample: Custom Content` I could not navigate to the page in the link. Pressing `enter` or `space` triggered the flip. I also noticed that in both Safari and Firefox I could not tab into the link, the focus stayed on the card. I hope this is clear and if you have any question feel free to call me.

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,316 @@
...
+      :host {
+        @apply block aspect-3/4;
+        --name: '';
+        --height: 480px;
+        height: var(--height);
+      }
+
+      .flip-card {
+        perspective: 100rem;
+      }
```
smfonseca, line 264: NIT: please add whitespace (also on L296 and L303)

van-nguyen-ht (approved):
I checked the link and everything is aligned with design now.

smfonseca (approved):

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -86,18 +86,20 @@ export default class SdFlipcard extends SolidElement {
   }
 
   private handleFrontKeydown(event: KeyboardEvent) {
-    if (event.code === 'Enter' || event.code === 'Space') {
+    console.log(event.target, this.front, this.front === event.target);
+    console.log('handleFrontKeydown');
+    console.log(this.back === event.target, event.type, event.view, event.ctrlKey);
```
mariohamann: Remember to remove console.log

Vahid1919: I think this was an older version :)

packages/components/src/components/flipcard/flipcard.stories.ts
Vahid1919: created the ticket here: #1175

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,337 @@
...
+   */
+  @property({ reflect: true }) activation: 'click-only' | 'click-and-hover' = 'click-and-hover';
+
+  /**
+   * Allows the flipcard to flip vertically or horizontally.
+   */
+  @property({ reflect: true, attribute: 'flip-direction' }) flipDirection: 'horizontal' | 'vertical' = 'horizontal';
+
+  /** Determines the variant of the front face of the flipcard. */
+  @property({ type: String, reflect: true })
```
mariohamann, line 52: Didn't we agree that we add an empty one as default?

packages/components/src/components/flipcard/flipcard.ts
```diff
@@ -0,0 +1,337 @@
...
+  frontVariant:
+    | 'primary'
+    | 'primary-100'
+    | 'gradient-light-top'
+    | 'gradient-light-bottom'
+    | 'gradient-dark-top'
+    | 'gradient-dark-bottom' = 'primary';
+
+  /** Determines the variant of the back face of the flipcard. */
+  @property({ type: String, reflect: true }) backVariant:
```
mariohamann, line 62: Didn't we agree that we add an empty one as default?

mariohamann: See https://github.com/solid-design-system/solid/pull/1121#issuecomment-2146984828

mariohamann (approved):

Issue #664, opened 2023-12-19
feat: ✨ Create dedicated spacing tokens for 1px


Vahid1919:
## User Story
As a Developer of the Solid Design System, I want to enhance the efficiency and maintainability of our codebase by introducing a dedicated spacing token for 1px in the Solid Design System. This will allow me to use a standardised value directly, eliminating the need to rely on arbitrary values which create unnecessary bloat.

Example: We are currently using arbitrary values in sd-teaser #663, and is currently in the development version of sd-tab. 

### Suggested Solution
Create a style(spacing) token for 1px (and more granular values -> could be discussed in refinement) 

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

coraliefeil:
@van-nguyen-ht you can tackle this one. @yoezlem is fine with it.

van-nguyen-ht:
I created a PR and assigned you @Vahid1919 :)

Vahid1919:
This is resolved by #1152

PR #1152, opened 2024-06-12
ci(token): add spacing.0,25 token of 1px

van-nguyen-ht:
## Description:
following #664
spacing token for 1px need to be added

## Definition of Reviewable:
- [x] update Figma documentation


mariohamann:
What do you think about calling it just "px" instead of 0.25? THis feels way to abstract.

Vahid1919:
> What do you think about calling it just "px" instead of 0.25? THis feels way to abstract.

Agreed. it does not seem consistent with the other values for spacing tokens. It seems like it should be 0.25 instead of "px"

van-nguyen-ht:
@Vahid1919 oh I just copy the same name from Tailwind, I have no problem of it naming 0.25 :)

yoezlem (approved):

Vahid1919 (approved):

PR #1156, opened 2024-06-18
fix: add utilities to exports

mariohamann:
Enables to import utilities. This solves to easily import e. g. `autocomplete-config` without having to point to the direct path via `node_modules/@solid-design-system/...`.

Everything was tested locally:

CleanShot 2024-06-18 at 11 05 25@2x

## Definition of Reviewable:

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -19,12 +19,20 @@
       "import": "./dist/package/components/*",
       "types": "./dist/package/solid-components.d.ts"
     },
+    "./unversioned-utilities/*": {
+      "import": "./dist/package/utilities/*",
+      "types": "./dist/package/solid-components.d.ts"
+    },
     "./versioned": {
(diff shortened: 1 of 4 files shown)


packages/components/src/utilities/autocomplete-config.ts
mariohamann, line 1: This auto-imports the popup, which is neede for the autocomplete.

packages/components/package.json
```diff
@@ -19,6 +19,10 @@
       "import": "./dist/package/components/*",
       "types": "./dist/package/solid-components.d.ts"
     },
+    "./unversioned-utilities/*": {
```
mariohamann, line 22: Info: Exports will be optimized and simplified when the architecture is optimized.

smfonseca (approved):

Issue #270, opened 2023-07-26
feat: ✨ add sd-audio


karlbaumhauer:
As a stakeholder of the Solid Design System, I want to utilize the new `sd-audio` component to provide audio content in a more appealing and dynamic way, so that end-users can engage with the content more effectively and have a better user experience.

## Documentation & Hand-off
https://www.figma.com/file/cHffrwo9yDeKpqsPcYN8R7/sd-audio-V.1.0.0?type=design&node-id=3112-19985&mode=design

#### Props
| Status | Name | Type | Default | Description |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| <ul><li>[x] </li></ul>  | `inverted` | `boolean` | `false` | inverts the colors for use on primary background |
| <ul><li>[x] </li></ul>  | `reversedLayout` | `boolean` | `false` | positions the controls below the timeline |
| <ul><li>[x] </li></ul>  | `animated` | `boolean` | `false` | activates the wave animation |

#### Parts (besides the shoelace parts)
  - [ ] every other HTML element (depending on the HTML structure) => NOTE: if highlighting the parts in the `parts` story does not work out of the box, do not spend more effort on it but open a bug ticket to tackle it later.
#### Slots
  - [x] `default` - place for the audio player
  - [x] `transscript` - by default hidden slot to provide the transcript (only if the <track> solution doesn't work).
  - [x] `play/pause` - provides a slot to replace play and pause icon/button?!
#### Stories 
  - [x] `default` - displays the default state of the component
  - [x] `slots` - displays all the slots
  - [x] `reversed layout` - displays the component with reversed control order
  - [x] `parts` - all the parts
  - [x] `backgrounds` - shows the player on 3 different backgrounds (as in Figma)

#### Samples
  - [x] `large info` - shows a player with complex content
  - [x] `transcript` - shows a player with added transcript
  - [x] `teaser` - shows the player with a teaser

### Open questions towards development
- [x] Do we need the speed toggle as property to be able to set it programmatically? => nope
- [x] The same with the transscript? => nope

### Open Questions towards design
- [x] [standard HTML player](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_audio) has a "download" feature implemented into the 3-dot-menu (next to speed). Do we use this as well? 
-> IMO we cannot offer a download functionality due to legal / license reasons. Since we did not discuss it earlier and it hasn’t bee requested yet ... I would skip that.  
[...]

PR #1194, opened 2024-07-07
feat: sd-audio

smfonseca:
## Description:

This PR addresses [ticket #270](https://github.com/solid-design-system/solid/issues/270) and is still a work in progress.
If someone in the team wants to continue with it while I'm away, here's the list of remaining todos:

- Add tests
- Add migration guide
- Fix bug with the hidden volume slider in Safari
- Clarify initial idea for implementation of transcript vtt file (it's not supported by the audio element)
- Add "samples" stories
- `play / pause` -> primary button with overriden size and radius

## Definition of Reviewable:


packages/components/src/components/audio/audio.stories.ts
```diff
@@ -0,0 +1,366 @@
...
+export default {
+  title: 'Components/sd-audio',
+  component: 'sd-audio',
+  args,
+  argTypes,
+  parameters: { ...parameters },
+  decorators: [withActions] as any
+};
+
+const audioElementScript = html`<script>
```
karlbaumhauer, line 19: do we really need all of that code in Storybook? looks like our component is not able to be used without all of that... :/

image

packages/components/src/components/audio/audio.stories.ts
karlbaumhauer, line 1: for some reason I have to reload the whole page to make the player run. Especially as the animation is running but the player doesn't, it looks broken

https://github.com/user-attachments/assets/02a9639d-c031-40d0-99e9-d18110190c8c

packages/components/src/components/audio/audio.stories.ts
```diff
@@ -0,0 +1,366 @@
...
+    return html`
+      <div id="sd-audio-example" className="p-0">
+        ${generateTemplate({
+          args,
+          constants: [
+            audioConstant,
+            {
+              type: 'slot',
+              name: 'transcript',
+              value: '<div>Show audio transcript</div>'
```
karlbaumhauer, line 94: can we add our transcript here? Files should be available in our teams channel (under files > sample audio > sds-song_transcript.vtt

packages/components/src/components/audio/wave.ts
karlbaumhauer, line 1: When I first visited the chromatic deployment, the animation was not starting. Now it only happens when switching between animated and normal WHILE the player is running. Don't know if this is a use-case but even if we cannot fix it, we maybe name it.

packages/placeholders/src/audio/sds-song_transcript.vtt
karlbaumhauer, line 1: as we agreed not to have the transcript as "real" audio transcript but only as text, why do we hav this file in the placeholders?

karlbaumhauer: or, where is it used?

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,603 @@
...
+  /** Reverts the order of the audio controls and timestamps */
+  @property({ type: Boolean, reflect: true }) reversedLayout = false;
+
+  /** Shows or hides the timestamps */
+  @property({ type: Boolean, reflect: true }) hideTimestamps = false;
+
+  /** Enables or disables the wave animation when the audio is playing */
+  @property({ type: Boolean, reflect: true }) animated = false;
+
+  //** Inverts the colors of the component elements and adds a bg-primary background to the container */
```
karlbaumhauer, line 55:
```suggestion
  /** Inverts the colors of the component elements and adds a bg-primary background to the container */
```

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,603 @@
...
+  @property({ type: Boolean, reflect: true }) inverted = false;
+
+  //** Background variants of the component container */
```
karlbaumhauer, line 58:
```suggestion
  /** Background variants of the component container */
```

karlbaumhauer (changes requested):

yoezlem:
When I have the animated version of the player on and it is paused, the entire animation is interrupted. In the old library, the animation continues both when animated is set to true and after clicking pause.

image

yoezlem:
Shouldn't the volume slider be all the way to the left when muted?

image

coraliefeil:
Pls reassign for another review. THX ;)

packages/components/src/components/audio/audio.stories.ts
mariohamann: Good point. My expectation would be, that default HTML audio would be wired up automatically.

MartaPintoTeixeira:
@yoezlem I was reading your comment regarding the animation of the Audio when paused.
"When I have the animated version of the player on and it is paused, the entire animation is interrupted. In the old library, the animation continues both when animated is set to true and after clicking pause."

I would like to ask if there is any reason why you want to keep the animation (apart from the fact that it was like this in the previous component). 

I ask this because while aligning with @coraliefeil she mentioned that the team agreed on stopping the animation when paused "Animation should be static until the audio-track is played and it should stop when paused." as from a UX perspective we think it makes more sense.

Please let us know your thoughts on the topic so we can decide on the result.

yoezlem:
> @yoezlem I was reading your comment regarding the animation of the Audio when paused. "When I have the animated version of the player on and it is paused, the entire animation is interrupted. In the old library, the animation continues both when animated is set to true and after clicking pause."
> 
> I would like to ask if there is any reason why you want to keep the animation (apart from the fact that it was like this in the previous component).
> 
> I ask this because while aligning with @coraliefeil she mentioned that the team agreed on stopping the animation when paused "Animation should be static until the audio-track is played and it should stop when paused." as from a UX perspective we think it makes more sense.
> 
> Please let us know your thoughts on the topic so we can decide on the result.

There isn't a specific reason why I want this, but when I compare it with the old library, we initially had this behavior implemented. That’s why I assumed we would also include it in the new component. However, if you're against it from a design perspective and have discussed it, that's fine with me, and we can leave it out.
https://component-library.dev.fe.union-investment.de/integration/storybook/index.html?path=/story/components-audio-player--animated

yoezlem:
LGTM approved

yoezlem (approved):

packages/components/src/components/audio/wave.ts
karlbaumhauer: This behavior is still visible. As soon as I started the player, all the props like (animated, timestamp etc) do not work anymore.

karlbaumhauer: actually they only work once in general. Is this related to Chromatic deployment?

packages/components/src/components/audio/audio.ts
```diff
@@ -399,52 +387,42 @@ export default class SdAudio extends SolidElement {
...
+              ? 'text-white focus-visible:focus-outline-inverted'
+              : 'text-primary focus-visible:focus-outline'
+          )}
+          part="volume"
+          aria-label=${this.localize.term('mute')}
+          tabindex="0"
+          @click=${this.toggleMute}
+          @keydown=${this.toggleMuteKeydown}
+        >
+          <sd-icon class="w-6 h-6" name=${this.isMuted ? 'mute' : 'volume'} library="system"></sd-icon>
```
karlbaumhauer, line 424: what was the latest agreement on the slider? It is completely gone in the feature branch deployments.

karlbaumhauer (commented):
@smfonseca pls also check the list in the epic ticket #270 

a few checkboxes are still open.

packages/components/src/components/audio/wave.ts
smfonseca: I talked with @mariohamann and this seems to be a storybook bug. We may have to update one of our dependencies to have it fixed. I will open a ticket.

But in the meantime, the `animated` prop is working properly.

packages/components/src/components/audio/audio.ts
smfonseca: The volume slider was presenting some ux and implementation issues so we decided to scrap it for now and go with simpler version. If we see it is something our users need, we will find a way to include it in a future iteration.

smfonseca:
@yoezlem @MartaPintoTeixeira I've implemented again the feature for the animation to start and stop according the the player state.

karlbaumhauer (approved):

packages/components/src/components/audio/audio.stories.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Use the `inverted` attribute to set the audio’s variant up side down.
+ */
+export const Inverted = {
+  render: () => {
+    return html`
+      <div class="bg-primary">
+        <sd-audio inverted>
+          <audio slot="default" src="./placeholders/audio/sds-song.mp3" preload="metadata"></audio>
```
mariohamann, line 71:
```suggestion
          <audio src="./placeholders/audio/sds-song.mp3" preload="metadata"></audio>
```


There shouldn't be a "default" slot named here? It's just... the main slot.

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+@customElement('sd-audio')
+export default class SdAudio extends SolidElement {
+  private readonly localize = new LocalizeController(this);
+
+  private readonly hasSlotController = new HasSlotController(this, 'transcript');
+
+  /** Reverses the order of the audio controls and timestamps */
+  @property({ type: Boolean, reflect: true, attribute: 'reversed-layout' }) reversedLayout = false;
+
+  /** Shows or hides the timestamps */
```
mariohamann, line 48:
```suggestion
  /** Hides the timestamps */
```

Not sure, if we do it like this in other components.

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+  @property({ type: Boolean, reflect: true, attribute: 'hide-timestamps' }) hideTimestamps = false;
+
+  /** Enables or disables the wave animation */
```
mariohamann, line 51:
```suggestion
  /** Enables the wave animation */
```

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+  @property({ type: Boolean, reflect: true }) animated = false;
+
+  /** Inverts the colors of the component elements in darker backgrounds */
```
mariohamann, line 54:
```suggestion
  /** Inverts the colors of the component. */
```

It's not changing "on dark backgrounds".

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+        !this.animated && 'relative',
+        this.animated && !this.reversedLayout && 'absolute top-0 left-0 w-full',
+        this.reversedLayout ? 'mt-2' : 'mb-2'
+      )}
+      part="audio-controls"
+    >
+      <button
+        class=${cx(
+          'audio-player__playback-speed justify-self-start text-base font-bold hover:cursor-pointer hover:text-primary-500',
+          this.inverted ? 'text-white focus-visible:focus-outline-inverted' : 'text-primary focus-visible:focus-outline'
```
mariohamann, line 363: Couldn't you use `sd-interactive` here?

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+      </div>
+    </div>`;
+
+    return html`
+      <div
+        class=${cx('w-full flex p-12 relative', this.reversedLayout ? 'flex-col-reverse' : 'flex-col')}
+        aria-label=${this.localize.term('audioPlayer')}
+        part="audio-player"
+      >
+        <slot name="default" @slotchange="${this.updateAudioDuration}"></slot>
```
mariohamann, line 464:
```suggestion
        <slot name="default" @slotchange="${this.updateAudioDuration}"></slot>
```

This should be just "<slot></slot>"! And then query e. g. for the first element inside the slot.

packages/components/src/components/audio/audio.ts
```diff
@@ -0,0 +1,539 @@
...
+        @click=${!this.isPlaying ? this.playAudio : this.pauseAudio}
+        aria-label="${this.isPlaying ? this.localize.term('pauseAudio') : this.localize.term('playAudio')}"
+        tabindex="0"
+      >
+        ${this.isPlaying
+          ? html`<slot name="pause-icon">
+              <sd-icon
+                name="pause"
+                library="system"
+                class=${cx('w-6 h-6', this.inverted ? 'text-primary' : 'text-white')}
```
mariohamann, line 389: You could set the color just on the button above, so you don't have to redefine it twice. This furthermore would make it easier if people slot other elements as well. Same for all other elements.

mariohamann (changes requested):

MartaPintoTeixeira:
@smfonseca I have just aligned with Coralie regarding the behaviour when we pause the animation and we agreed that makes more sense to have the "sound waves" stopping instead of disappearing. Let me know if you have questions.

packages/components/src/components/audio/audio.stories.ts
smfonseca: Updated

packages/components/src/components/audio/audio.ts
smfonseca: Updated

packages/components/src/components/audio/audio.ts
smfonseca: Updated

packages/components/src/components/audio/audio.ts
smfonseca: Updated

packages/components/src/components/audio/audio.ts
smfonseca: Updated

MartaPintoTeixeira:
As spoken:
- for the animation: please chose a 1st preview that is more pleasant visually (something similar to what we have in sec.4 could work).
- on hover: please make sure the play button behaves the same way as a tertiary button with icon.
- speed and volume: should be "interactive" components if the component has issues we need to create a fix ticket or override it please

MartaPintoTeixeira:
Please add "Related templates: Audio" to docs.
And create the respective template.

MartaPintoTeixeira:
@coraliefeil as spoken the play button behaves now as a "button". Looking at mouse hover in both default and inverted I think you almost don't realize the tones as changing. Let's check this together next week.

@smfonseca the rest of the changes you made look good.

coraliefeil:
> @coraliefeil as spoken the play button behaves now as a "button". Looking at mouse hover in both default and inverted I think you almost don't realize the tones as changing. Let's check this together next week.

Interaction states look as intended. No issues here. Let’s ship it <3

MartaPintoTeixeira (approved):

mariohamann (approved):

PR #1213, opened 2024-07-18
fix: solid faker seed for paragraphs and sentences

christophsaile:
## Definition of Reviewable:

### DIFF
--- a/packages/placeholders/README.md
+++ b/packages/placeholders/README.md
@@ -16,12 +16,6 @@ import SolidFaker from '@solid-design-system/placeholders';
 // Create an instance with an optional seed value as the only argument
 const faker = new SolidFaker(123);
 
-// Set the seed to any other number to provide a different set of consistent results
-faker.seed(321);
-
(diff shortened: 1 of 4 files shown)


christophsaile:
@solid-design-system/core-development it would be nice if someone could verify that everything is working. I tried different scenarios and all of them looked good but just to be sure :)

smfonseca (commented):
I've tested your changes and all seems to be fine, the only request I would have is that we move the script and its tests into an `utilities` folder just to keep everything organized.

mariohamann:
> I've tested your changes and all seems to be fine, the only request I would have is that we move the script and its tests into an `utilities` folder just to keep everything organized.

As this utility is in its own package, I would leave it as is, as it's the `src` of this package.

smfonseca:
I agree with your point @mariohamann but I'm also looking into the future where we may add other utilities and then it can start becoming unorganized. But this is something we can care about when the time arrives. Let's just move forward with the PR.

smfonseca (approved):

PR #1215, opened 2024-07-23
feat: 🎸 adding min and max posibility for date type of input

JasminaIlic:
## Description:
input type - date : min and max properties (of type type of string) was added so that type of input:Date can also use those properties.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/input/input.test.ts
+++ b/packages/components/src/components/input/input.test.ts
@@ -82,16 +82,23 @@ describe('<sd-input>', () => {
   });
 
   describe('value min & max', () => {
-    it('should react on value within of range form date', async () => {
+    it('should react with data-valid attribute on input field when date value is within min-max range of date', async () => {
       const el = await fixture<SdInput>(html`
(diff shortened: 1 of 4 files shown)


packages/components/src/components/input/input.test.ts
```diff
@@ -81,6 +81,32 @@ describe('<sd-input>', () => {
...
 
+  describe('value min & max', () => {
+    it('should react on value within of range form date', async () => {
+      const el = await fixture<SdInput>(html`
+        <sd-input type="date" value="2023-06-10" max="2024-06-10" spellcheck></sd-input>
+      `);
+      expect(el.hasAttribute('data-valid')).to.be.true;
+    });
+
+    it('should react on value out of range form date', async () => {
```
mariohamann, line 92:
```suggestion
    it('should react on value out of range form date', async () => {
```

Please describe in the test, that this is about invalidity and the max value.

mariohamann (approved):

PR #1220, opened 2024-07-26
fix: sd-tooltip self closing with click focus trigger (close #1211)

### LINKED ISSUE #1211: fix: 🤔 sd-tooltip closing itself on first open
## Current behavior
If the user clicks on the trigger for the tooltip, the tooltip opens and immediatly closes itself again. 
The user then has to open the tooltip a second time. 

## Expected behavior
The tooltip stays open after a click on its trigger.

## Steps to reproduce
1. Go to https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/story/components-sd-tooltip--default&args=open-attr:!false
2. Open the tooltip

## Technical Information
This only seems to happen if the _trigger_ attribute is set to "click focus".

A workaround for us was to set the trigger to "click" only, but then the user has to close the tooltip via click every time instead of the tooltip closing itself when loosing focus. So this is not really an option.

## DoR
- [x] Item has business value

## DoD

smfonseca:
## Description:
If the tooltip trigger is set to `click focus`, the focus event will be triggered first and the attached logic will run. Afterwards, the logic of the `click`event will also run, meaning that the tooltip will open and close immediately. In this PR we add a flag that will control which of the event handle should be executed.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/tooltip/tooltip.test.ts
+++ b/packages/components/src/components/tooltip/tooltip.test.ts
@@ -158,4 +158,61 @@ describe('<sd-tooltip>', () => {
 
     expect(getComputedStyle(popup.popup).pointerEvents).to.equal('none');
   });
+
+  it('should toggle the tooltip when clicking the trigger', async () => {
+    const el = await fixture<SdTooltip>(html`
+      <sd-tooltip content="This is a tooltip">
+        <sd-button>Click Me</sd-button>
+      </sd-tooltip>
+    `);
+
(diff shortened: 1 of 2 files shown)


mariohamann:
Could we have a test for that?

smfonseca:
> Could we have a test for that?

@mariohamann tests added, please review again when you can.
Thank you

mariohamann (approved):

PR #1231, opened 2024-08-01
feat: sd-accordion-group docs [skip chromatic]

smfonseca:
## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/accordion-group/accordion-group.stories.ts
+++ b/packages/components/src/components/accordion-group/accordion-group.stories.ts
@@ -2,8 +2,6 @@
 import '../../solid-components';
 import { html } from 'lit-html';
 import { storybookDefaults, storybookTemplate } from '../../../scripts/storybook/helper';
-import { userEvent } from '@storybook/test';
-import { waitUntil } from '@open-wc/testing-helpers';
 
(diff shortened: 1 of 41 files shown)


packages/components/src/components/accordion-group/accordion-group.stories.ts
```diff
@@ -34,124 +39,49 @@ export const Default = {
...
-  }
+export const SampleWithPrimaryBackground = {
+  name: 'Sample with primary background',
+  render: () => html`
+    <sd-accordion-group class="bg-primary-100 p-8">
+      <sd-accordion summary="Accordion 1">Sample content</sd-accordion
+      ><sd-accordion summary="Accordion 2">Sample content</sd-accordion
+      ><sd-accordion summary="Accordion 3">Sample content</sd-accordion>
+    </sd-accordion-group>
+  `
```
mariohamann, line 86: That's a weird one. @coraliefeil Why do we document this? This feels more like a UX hint, which could be done verbally but having this as a sample or explicitly visually documented feels... weird? (Especially: Why only for this, but not for others?)

smfonseca: Moved to `templates` as discussed in the sync with @coraliefeil and @MartaPintoTeixeira.

Issue #1131, opened 2024-06-06
feat: ✨ refactor monorepo by extracting docs


mariohamann:
## User Story
As a maintainer of the Solid Design System, I would like to have a more scalable and robust infrastructure, so that we're prepared for upcoming features and optimized documentation. Furthermore we want to upgrade our understanding from `Solid Components` to `Solid Design System` throughout our docs.

### Suggested Solution
1. We move Storybook including Vite and all needed stories out of `/components` into a separate package `/docs`, similar to what [Synergy (docs)](https://github.com/synergy-design-system/synergy-design-system/tree/main/packages/docs) does.
2. We introduce `packages/docs/stories/packages` as in [Synergy](https://github.com/synergy-design-system/synergy-design-system/tree/main/packages/docs/stories/packages) for all our packages, including CHANGELOGs and READMEs for every package.
3. We adapt CI-/CD-Pipelines in Azure, so that Storybook is in a seperate `\docs` folder, independent from versioning, e. 4. `https://solid-design-system.fe.union-investment.de/docs`. Global versions for the whole Storybook will then be removed, e. g. in the initial headline `Introduction of Solid Components v3.6.0`. Instead we will brand the Storybook to `Introduction of Solid Design System`.
5. Regarding the latter we should throughout the repo in READMEs etc. if they fit to what they represent.

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Implementation works successfully on `feature` branch

PR #1240, opened 2024-08-05
chore: fix playwright on CI [skip chromatic]

mariohamann:


Vahid1919 (approved):

mariohamann:
Thanks for the reviews, but this still not working. 😆

mariohamann:
Quick fixes apparently don't work. We should handle this in https://github.com/solid-design-system/solid/issues/1131

PR #1244, opened 2024-08-05
docs: sd-dialog docs [skip chromatic] 

smfonseca:
## Definition of Reviewable:


yoezlem (changes requested):
Seems like not in sync at all?

In Figma we have: 

- Variants
- No close button
- Queries
- Adding a backdrop
- Slots

Please get in touch with @coraliefeil or @MartaPintoTeixeira

smfonseca:
> Seems like not in sync at all?
> 
> In Figma we have:
> 
> * Variants
> * No close button
> * Queries
> * Adding a backdrop
> * Slots
> 
> Please get in touch with @coraliefeil or @MartaPintoTeixeira

There are some components which are ahead of design. This means that I've already started rewriting the documentation from dev side and when design starts working on it we align both sides. This is one of those cases. But I can already say that`queries` and `adding a backdrop` will not feature in storybook since they are only relevant for figma.

coraliefeil:
See comments on Figma.

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -1,20 +1,25 @@
...
 
 /**
- * **Disclaimer** : Input-based controls for `sd-dialog` are unavailable on this Docs page due to a [Storybook bug](https://github.com/solid-design-system/solid/issues/702). Please use the individual feature stories to access these controls.
+ *
+ * Appears over other content. It requires an interaction from the user before they can return to whatever is underneath.
+ *
+ *  **Related templates**:
+ * - [Dialog with scrollable content](?path=/docs/templates-dialog-with-scrollable-content--docs)
+ *
  */
```
coraliefeil, line 18: - Use the `headline` slot to add a headline.
- Use the `default` slot to add main content.  Default slot is always scrollable.
- Use the `footer` slot to add action elements. Footer slot is always fixed.

smfonseca: Updated

smfonseca:
@coraliefeil I've addressed the comments you've left and added the first template. Once I'm back I will add the remaining samples.

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -57,271 +66,60 @@ export const Default = {
...
+export const Open = {
+  name: 'Open',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog id="dialog" open>
+        <h4 slot="headline" class="sd-headline sd-headline-3xl">Lorem ipsum</h4>
+        <p class="sd-paragraph">
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nibh justo ullamcorper odio tempor molestie phasellus
+          dui vel id. Velit in sed
+        </p>
```
coraliefeil, line 81: Pls show primary button in footer slot on the right-hand side.

packages/components/src/templates/dialog-samples.stories.ts
```diff
@@ -0,0 +1,129 @@
...
+export const Default = {
+  name: 'Dialog with accordion',
+  render: () => html`
+    <style>
+      sd-dialog::part(footer) {
+        margin-left: 0;
+      }
+    </style>
+    <div class="min-h-[100vh]">
+      <sd-dialog id="dialog" open>
```
coraliefeil, line 32: In this use case the close button needs to be omitted because users have to click the primary button in order to close the dialog overlay.

packages/components/src/templates/dialog-samples.stories.ts
```diff
@@ -0,0 +1,129 @@
...
+            </sd-accordion>
+          </sd-accordion-group>
+        </div>
+
+        <div slot="footer" class="flex flex-col w-full space-y-5">
+          <sd-button variant="primary" size="lg" type="button"> Accept terms of use </sd-button>
+          <sd-button variant="secondary" size="lg" type="button"> Download terms of use as PDF </sd-button>
+          <sd-button variant="secondary" size="lg" type="button"> Exit IK portal </sd-button>
+          <p class="sd-paragraph text-center">
+            If you have any questions or concerns, <sd-link href="#">contact us</sd-link>.
```
coraliefeil, line 123: text color black should be deemphasized. pls use text.neutral-700 instead of text.black

packages/components/src/templates/dialog-samples.stories.ts
coraliefeil: I’ve updated the content ... since Storybook is public we need to make sure we only show public information

[See Figma ](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=2365-41319&t=2cNoGQ6ofKzqsyrT-11)

coraliefeil: How can we show the responsive behavior here?

coraliefeil:
See comments. Pls reassign when adjusted.

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -1,20 +1,29 @@
...
- * **Disclaimer** : Input-based controls for `sd-dialog` are unavailable on this Docs page due to a [Storybook bug](https://github.com/solid-design-system/solid/issues/702). Please use the individual feature stories to access these controls.
+ *
+ * Appears over other content. It requires an interaction from the user before they can return to whatever is underneath.
+ *
+ * - Use the `headline` slot to add a headline.
+ * - Use the `default` slot to add main content. <br /> Default slot is always scrollable.
+ * - Use the `footer` slot to add action elements. <br /> Footer slot is always fixed.
+ *
+ *  **Related templates**:
+ * - [Dialog samples](?path=/docs/templates-dialog-samples--docs)
```
yoezlem, line 19: Please Change "samples" to "dialog"

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -24,17 +33,17 @@ export default {
...
       type: 'slot',
       name: 'headline',
-      value: `<h4 slot="headline" class="sd-headline sd-headline--size-3xl leading-tight">Lorem Ipsum</h4>`
+      value: `<div slot="headline" class="slot slot--border slot--text h-16">Headline slot</div>`
     },
     {
       type: 'slot',
       name: 'footer',
-      value: `<sd-button slot="footer" class="w-full">Label</sd-button>`
+      value: `<div slot="footer" class="slot slot--border slot--text h-16 w-full">Footer slot</div>`
```
yoezlem, line 46: Please change the with of footer like in figma: 
image

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -57,271 +66,60 @@ export const Default = {
...
-    </div>`;
-  }
+export const Open = {
+  name: 'Open',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog id="dialog" open>
+        <h4 slot="headline" class="sd-headline sd-headline-3xl">Lorem ipsum</h4>
+        <p class="sd-paragraph">
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nibh justo ullamcorper odio tempor molestie phasellus
```
yoezlem, line 79: Its only a small thing, but in design its:
`Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam.`

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -57,271 +66,60 @@ export const Default = {
...
-      </div>
-    `;
-  }
+export const Headline = {
+  name: 'Headline',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog id="dialog" open headline="Headline">
+        <p class="sd-paragraph">
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nibh justo ullamcorper odio tempor molestie phasellus
```
yoezlem, line 97: Please adjust the text to match the Figma file
`Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam.`

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: Updated.

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: The footer slot is right aligned, therefore in my opinion the update should be done in Figma. 
FYI @coraliefeil

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: Updated.

packages/components/src/templates/dialog-samples.stories.ts
smfonseca: Viewport addon to display different viewports seems to not be available in the docs page. Did some research and could not find anything that we can plug n play to have it working. Need to discuss it with the team to check how we could approach it.

packages/components/src/templates/dialog.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+};
+
+/**
+ *
+ */
+
+export const Default = {
+  name: 'Dialog with Button Group',
```
coraliefeil, line 21: How about the mobile screen that shows up automatically when the viewport is smaller than 414px.

It has smaller headline and paddings and the buttons are streched and vertically stacked.

Could we show that in the storybook template somehow?
I cannot switch to a mobile view in the templates.

packages/components/src/templates/dialog.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+          <sd-button variant="secondary" size="lg" type="button"> Maybe later </sd-button>
+          <sd-button variant="primary" size="lg" type="button"> Cancel </sd-button>
+        </div>
+      </sd-dialog>
+    </div>
+  `
+};
+
+export const Scrollable = {
+  name: 'Dialog with scrollable content',
```
coraliefeil, line 42: How about the mobile screen that shows up automatically when the viewport is smaller than 414px.
It has smaller headline and paddings and the buttons are streched and vertically stacked.

Could we show that in the storybook template somehow?
I cannot switch to a mobile view in the templates.

coraliefeil: Pls increase the frame height to sth like 750px height and reduce the height of the dialog itself so that there is minium padding-top and padding-bottom (48px)
Bildschirmfoto 2024-08-28 um 10 50 45

packages/components/src/templates/dialog.stories.ts
```diff
@@ -0,0 +1,150 @@
...
+              <sd-link href="https://ui-link.de/richtlinien">https://ui-link.de/richtlinien</sd-link>. The respective
+              fund-issuing company may decide at any time to revoke any arrangements it may have made for the
+              distribution of shares of a fund and/or share classes of a fund in a member state other than its home
+              member state, subject to the conditions of Art. 93a of Directive 2009/65/EC and Art. 32a of Directive
+              2011/61/EU.
+            </p>
+          </div>
+        </sd-scrollable>
+        <div slot="footer" class="flex justify-end w-full space-x-2">
+          <sd-button variant="primary" size="lg" type="button"> Accept terms of use </sd-button>
```
coraliefeil, line 137: Pls change order - The primary button in a group always shows up on the right-hand side:
Bildschirmfoto 2024-08-28 um 10 52 41

coraliefeil:
Thanks @smfonseca pls the docs are fine. Pls see comments on tamplates.

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: After a chat with @mariohamann we noticed that there is bug in the footer styling. I've created a [ticket](https://github.com/solid-design-system/solid/issues/1325) to address it.
@coraliefeil we can keep figma as it is and will update storybook once the bugfix is addressed. 
FYI @yoezlem

packages/components/src/templates/dialog.stories.ts
smfonseca: It's not possible to create stories specifically to show responsive behaviour, this needs to be done by resizing the window. I've made some changes to the template to overcome a [bug](https://github.com/solid-design-system/solid/issues/1325) we discovered in the dialog

packages/components/src/templates/dialog.stories.ts
smfonseca: See previous comment regarding responsive behaviour.
I've updated the frame size and as best as possible.

packages/components/src/templates/dialog.stories.ts
smfonseca: Updated

coraliefeil:
Approved. Thanks a lot <3

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -57,271 +66,61 @@ export const Default = {
...
+export const NoCloseButton = {
+  name: 'No Close Button',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog id="dialog" headline="Headline" open no-close-button>
+        <p class="sd-paragraph">
+          Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et
+          dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam.
+        </p>
+        <sd-button slot="footer">Button</sd-button>
```
Vahid1919, line 122: @smfonseca @coraliefeil  It seems a little strange that we suggest our users to make the button a close button to follow ARIA principles, but our example does not do that. I noticed that originally it was a close button, did we decide to move away from it on purpose? --**specifically for the no close button story**
image

smfonseca: Good point @Vahid1919, thank you. 
@coraliefeil can you please update in figma? I've updated storybook.

coraliefeil: Nope. Would switch it back to just "Button" ... it’s about the action button closing the dialog ... could be sth like "Accept Terms and Conditions" or the like ... "Close" will never appear like that. It’s about a message the user has to accept or confirm so the overlay blocking the interface can close.

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -1,20 +1,29 @@
...
 const { generateTemplate } = storybookTemplate('sd-dialog');
 
 /**
- * **Disclaimer** : Input-based controls for `sd-dialog` are unavailable on this Docs page due to a [Storybook bug](https://github.com/solid-design-system/solid/issues/702). Please use the individual feature stories to access these controls.
+ *
+ * Appears over other content. It requires an interaction from the user before they can return to whatever is underneath.
+ *
+ * - Use the `headline` slot to add a headline.
+ * - Use the `default` slot to add main content. <br /> Default slot is always scrollable.
+ * - Use the `footer` slot to add action elements. <br /> Footer slot is always fixed.
```
yoezlem, line 16: We should decide whether to write the docs under the example in Figma or not. Right now, it’s not consistent—sometimes we have it in Storybook, and sometimes we don’t.

FYI @coraliefeil

packages/components/src/components/dialog/dialog.stories.ts
Vahid1919: @coraliefeil @smfonseca I see. I would be fine with keeping it as "Button" or "Accept Terms and Conditions". The main part is maintaining the closing behavior.

We should perhaps me more precise in our documentation in this case. Instead of: "you should always include a close button" maybe something like: "You should always include an action button that allows users to close the dialog."

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: We unfortunately cannot move the docs under the examples in storybook. @coraliefeil is this something you can update on your side?

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -1,20 +1,29 @@
...
 import { withActions } from '@storybook/addon-actions/decorator';
 
 const { overrideArgs } = storybookHelpers('sd-dialog');
 const { argTypes, parameters } = storybookDefaults('sd-dialog');
 const { generateTemplate } = storybookTemplate('sd-dialog');
 
 /**
- * **Disclaimer** : Input-based controls for `sd-dialog` are unavailable on this Docs page due to a [Storybook bug](https://github.com/solid-design-system/solid/issues/702). Please use the individual feature stories to access these controls.
+ *
+ * Appears over other content. It requires an interaction from the user before they can return to whatever is underneath.
```
MartaPintoTeixeira, line 12:
```suggestion
 * Used over other content. It requires an interaction from the user before they can return to whatever is underneath.

File: packages/components/src/components/dialog/dialog.stories.ts (hunk 12 above)
Comment (smfonseca) [reply packages/components/src/components/dialog/dialog.stories.ts]: I like your suggestion @Vahid1919. What do you think @coraliefeil? Can we update the description to something like what @Vahid1919 mentioned?

File: packages/components/src/components/dialog/dialog.stories.ts [hunk 15]
```diff
@@ -57,271 +67,73 @@ export const Default = {
...
+export const NoCloseButton = {
+  name: 'No Close Button',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog id="dialog" headline="Headline" open no-close-button>
+        <p class="sd-paragraph">
+          Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et
+          dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam.
+        </p>
+        <sd-button class="close-button" slot="footer" variant="primary">Accept Terms and Conditions</sd-button>
```
MartaPintoTeixeira, line 123: Please replace button copy with "Close" as it is in FIGMA

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -1,20 +1,30 @@
...
- * **Disclaimer** : Input-based controls for `sd-dialog` are unavailable on this Docs page due to a [Storybook bug](https://github.com/solid-design-system/solid/issues/702). Please use the individual feature stories to access these controls.
+ *
+ * Used over other content. It requires an interaction from the user before they can return to whatever is underneath.
+ *
+ * - Use the `headline` slot to add a headline.
+ * - Use the `default` slot to add main content. <br /> Default slot is always scrollable.
+ * - Use the `footer` slot to add action elements. <br /> Footer slot is always fixed.
+ *
+ *
+ *  **Related templates**:
```
MartaPintoTeixeira, line 19: For the 1st preview of Dialog can we make it more similar to what we have in FIGMA?
"Footer slot" is the same size as other slots in FIGMA

packages/components/src/components/dialog/dialog.stories.ts
coraliefeil: No we keep it as is. We already aligned on that.

coraliefeil:
Approved. THX

yoezlem (approved):

Vahid1919 (approved):

packages/components/src/components/dialog/dialog.stories.ts
```diff
@@ -57,271 +67,73 @@ export const Default = {
...
-        </sd-dialog>
-      </div>
-    `;
-  }
+export const Open = {
+  name: 'Open',
+  render: () => html`
+    <div class="h-[40vh]">
+      <sd-dialog open>
+        <h4 slot="headline" class="sd-headline sd-headline-3xl">Lorem ipsum</h4>
```
karlbaumhauer, line 78:
```suggestion
        <h4 slot="headline" class="sd-headline sd-headline-3xl">Headline</h4>
```

packages/components/src/components/dialog/dialog.stories.ts
karlbaumhauer, line 123: I know there is an ongoing discussion on the content of the button but this needs to be fixed.

karlbaumhauer (changes requested):

packages/components/src/templates/dialog.stories.ts
```diff
@@ -0,0 +1,164 @@
...
+            found under “Notices and Complaints” at
+            <sd-link href="https://ui-link.de/richtlinien">https://ui-link.de/richtlinien</sd-link>. The respective
+            fund-issuing company may decide at any time to revoke any arrangements it may have made for the distribution
+            of shares of a fund and/or share classes of a fund in a member state other than its home member state,
+            subject to the conditions of Art. 93a of Directive 2009/65/EC and Art. 32a of Directive 2011/61/EU.
+          </p>
+        </div>
+      </sd-scrollable>
+      <div slot="footer" class="flex flex-col md:flex-row justify-end w-full gap-2">
+        <sd-button variant="secondary" size="lg" type="button" aria-label="Close">
```
karlbaumhauer, line 152: wrong button order compared to design. pls adjust.

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: @karlbaumhauer why do you suggest this change? It's approved by design to use lorem ipsum in this case.

packages/components/src/templates/dialog.stories.ts
smfonseca: Did you check the preview @karlbaumhauer? Because the order is aligned with design, the only situation where it's not it's actually on the smaller viewport.

packages/components/src/components/dialog/dialog.stories.ts
coraliefeil: Yes. THX

image

coraliefeil:
As shown in the call ... mobile order for stacked button is now fine ... primary action on top.

So approved (again) from my side.

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: After checking with @coraliefeil, we will not update.

packages/components/src/templates/dialog.stories.ts
smfonseca: Fixed for smaller viewports

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: Updated according to our sync

packages/components/src/components/dialog/dialog.stories.ts
smfonseca: This has been addressed

PR #1247, opened 2024-08-06
fix: 🤔 sd-carousel - numbered variant - show total clicks

### LINKED ISSUE #1191: fix: 🤔 sd-carousel - numbered variant - show total clicks
## Current behavior

For multiple slides per view:

The numbered variant does not show the total amount of clicks on the right: < 1 / 10 > (Immediate fix needed!)

## Expected behavior

The numbered variant shows the total amount of clicks on the right: < 1 / 10 >

Image

## Steps to reproduce

https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/story/components-sd-carousel--default&args=variant-attr:10;slides-per-page-attr:2;slides-per-move-attr:1&globals=outline:!true

## DoR
- [x] Item has business value

## DoD

Vahid1919:
## Description:

Closes  #1191. Updating scrolling logic to ensure that the page number calculation is accurate no matter the divisibility of slides-per-page and slides-per-move. 

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/carousel/carousel.stories.ts
+++ b/packages/components/src/components/carousel/carousel.stories.ts
@@ -16,11 +16,17 @@ export default {
       type: 'slot',
       name: 'default',
       value: `
+       
         <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 1</div></sd-carousel-item>
         <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 2</div></sd-carousel-item>
         <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 3</div></sd-carousel-item>
         <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 4</div></sd-carousel-item>
-        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 5</div></sd-carousel-item>`
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 5</div></sd-carousel-item>
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 6</div></sd-carousel-item>
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 7</div></sd-carousel-item>
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 8</div></sd-carousel-item>
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 9</div></sd-carousel-item>
+        <sd-carousel-item><div class="slot slot--border slot--text h-16">Default slot 10</div></sd-carousel-item>`
     }
   ]),
   argTypes,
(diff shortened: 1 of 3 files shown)


packages/components/src/components/carousel/carousel.ts
karlbaumhauer, line 1: @Vahid1919 are you sure this works as expected? I was able to configure some strange behavior again: 
image
IMO, no matter what we configure, it should never be 5 out of 4 right? I also tried to play around with your note you added but even when taking care on the numbers being divisible, I still hade 4 out of 3 clicks... :/

packages/components/src/components/carousel/carousel.ts
```diff
@@ -84,6 +84,12 @@ export default class SdCarousel extends SolidElement {
    */
```
karlbaumhauer, line 84: unfortunately I cannot tag line 56 here but I found a strange behavior in Storybook on the variant property: 
image
shouldn't this be a string property instead of number? maybe the term `number` is reserved here... :/

karlbaumhauer (changes requested):

packages/components/src/components/carousel/carousel.ts
Vahid1919: I accidentally introduced a bug when I was cleaning up my code. This should be working now.

packages/components/src/components/carousel/carousel.ts
karlbaumhauer: this works now.

packages/components/src/components/carousel/carousel.ts
karlbaumhauer: I think youre comment was towards my other conversation above right?

I will open up a new bug ticket as I found something else.

karlbaumhauer (approved):
approved from documentation perspective

coraliefeil:
Approved

yoezlem (approved):

packages/components/src/components/carousel/carousel.ts
```diff
@@ -342,7 +373,11 @@ export default class SdCarousel extends SolidElement {
       canSnap = Math.abs(previousIndex - this.slidesPerMove) % this.slidesPerMove === 0;
     }
 
-    this.goToSlide(previousIndex, behavior);
+    if (this.currentPage - 1 === 0 && this.loop) {
+      this.goToSlide(this.activeSlide - this.slidesPerPage, behavior);
+    } else {
+      this.goToSlide(previousIndex, behavior);
+    }
```
mariohamann, line 380: <video src="https://github.com/user-attachments/assets/cf0a48bf-02e1-4a04-bf46-1ddea036405e"></video>

This behaviour feels wrong.
- 5 slides, 2 per move, 2 per page
- Current: [4 5] → [3 4] → [1 2]
- Should: [4 5] → [2 3] → [1 2]

In the initial ticket ti says "it should always move the amount of slides per page as described if possible" – which would be currently violated.

mariohamann (changes requested):

packages/components/src/components/carousel/carousel.stories.ts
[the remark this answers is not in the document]
mariohamann: And sometimes Slide 2. On Safari.

Vahid1919:
Local tests: 
image

mariohamann (approved):

PR #1280, opened 2024-08-13
docs: sd-copyright docs [skip chromatic]

auroraVasconcelos:
## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/copyright/copyright.stories.ts
+++ b/packages/components/src/styles/copyright/copyright.stories.ts
@@ -6,7 +6,7 @@ const { overrideArgs } = storybookHelpers('sd-copyright');
 const { generateTemplate } = storybookTemplate('sd-copyright');
 
 /**
- * Can be displayed at the bottom of an image for example.
+ * Use the CSS property `--copyright` and the class `sd-copyright` in the parent of any element to set a copyright text.
  */
 export default {
   title: 'Styles/sd-copyright',
--- a/packages/components/src/styles/copyright/copyright.stories.ts
+++ b/packages/components/src/styles/copyright/copyright.stories.ts
@@ -1,9 +1,9 @@
 import '../../solid-components';
-import { html } from 'lit-html';
-import { storybookDefaults, storybookHelpers } from '../../../scripts/storybook/helper';
+import { storybookDefaults, storybookHelpers, storybookTemplate } from '../../../scripts/storybook/helper';
 
 const { argTypes, parameters } = storybookDefaults('sd-copyright');
 const { overrideArgs } = storybookHelpers('sd-copyright');
+const { generateTemplate } = storybookTemplate('sd-copyright');
 
 /**
  * Can be displayed at the bottom of an image for example.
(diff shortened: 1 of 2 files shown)


MartaPintoTeixeira (approved):
Looking good. In the future when possible we need to add a template with the Brandshape with Copyright variant.

mariohamann:
> Looking good. In the future when possible we need to add a template with the Brandshape with Copyright variant.

Could you please make a ticket for that?

packages/components/src/styles/copyright/copyright.stories.ts
```diff
@@ -29,17 +29,9 @@ export default {
...
-        templateContent: `<div class="%CLASSES% max-w-xl" style="--copyright: '${args['--copyright']}';">%SLOT%</div>`
-      },
-      args
-    });
-  }
+  name: 'Default',
+  render: () =>
+    html`<div class="sd-copyright max-w-xl" style="--copyright: '© 2024 Solid Design System';">
+      <img src="./placeholders/images/generic.jpg" alt="A generic placeholder jpg" class="aspect-video object-cover" />
+    </div>`
```
mariohamann, line 36: First story has to be interactive.

mariohamann: CleanShot 2024-08-19 at 07 07 44@2x

Currently broken.

yoezlem:
@coraliefeil @mariohamann 
it's quite difficult to read the copyright in the placeholder. Shouldn't we upload a darker image here and maybe consider having two types of font colors for the future?

coraliefeil:
> @coraliefeil @mariohamann it's quite difficult to read the copyright in the placeholder. Shouldn't we upload a darker image here and maybe consider having two types of font colors for the future?

@auroraVasconcelos has opened an issue for future implementation.

packages/components/src/styles/copyright/copyright.test.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+  },
+  render: (args: { [k: string]: any }) => {
+    return generateTemplate({
+      options: {
+        templateContent: `<div class="%CLASSES% max-w-xl" style="--copyright: '${args['--copyright']}';">%SLOT%</div>`
+      },
+      args
+    });
+  }
+};
```
mariohamann, line 45: I'm missing documentation of `--copyright` here. I know, that this already to questions in a team, there it would be good to explicitly mention it again.

Furthermore the information would help that you can wrap it around any element and that the text will show up at the right position.

So maybe something like:
> Wrap any element with the class and use the `--copyright` CSS custom property to set the text.

I'm pretty sure, that you will find a better solution, but you know which direction I'm thinking. :)

auroraVasconcelos: What about:

>Use the CSS property `--copyright` to set the text inside a component using the class `sd-copyright`.

mariohamann: That's not correct. The class with `sd-copyright` has to surround the element, that should get the copyright. E. g.

```html
<div class="sd-copyright" style="--copyright: '© 2024 Solid Design System';">
  <img src="whatever">
</div>
```

From the description you wrote it sounds more like the following should work:
```html
<img
  class="sd-copyright"
  style="--copyright: '© 2024 Solid Design System';"
  src="whatever"
>
```

But it doesn't.

auroraVasconcelos: > Use the CSS property `--copyright` and the class `sd-copyright` in the parent of any element to set a copyright text

yoezlem (approved):

PR #1285, opened 2024-08-13
docs: sd-tooltip docs [skip chromatic]

smfonseca:
## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/tooltip/tooltip.stories.ts
+++ b/packages/components/src/components/tooltip/tooltip.stories.ts
@@ -41,7 +41,6 @@ export default {
 
           #anchor--components-sd-tooltip--size .innerZoomElementWrapper,
           #anchor--components-sd-tooltip--placement .innerZoomElementWrapper,
-          #anchor--components-sd-tooltip--trigger .innerZoomElementWrapper,
           #anchor--components-sd-tooltip--hoist .innerZoomElementWrapper,
           .template-height {
             height: 110px;
@@ -49,7 +48,16 @@ export default {
 
           #anchor--components-sd-tooltip--placement .innerZoomElementWrapper {
             padding: 50px 0;
-            height: 300px;
+            height: 400px;
+          }
+
+          #anchor--components-sd-tooltip--trigger .innerZoomElementWrapper {
+            padding: 50px 0;
+            height: 200px;
+          }
+
+          #anchor--components-sd-tooltip--hoist .innerZoomElementWrapper {
+            height: 150px;
           }
         </style>
         ${story()}`
@@ -97,24 +105,42 @@ export const Size = {
  */
 export const Placement = {
   render: () => html`
-    <div class="grid grid-cols-3 gap-12">
+    <div class="grid md:grid-cols-3 gap-24">
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="top-start" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip
+          content="Top Start Positioning"
+          placement="top-start"
+          open
+          size="lg"
+          trigger="click focus"
+        ></sd-tooltip>
       </div>
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="top" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip content="Top Positioning" placement="top" open size="lg" trigger="click focus"></sd-tooltip>
       </div>
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="top-end" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip content="Top End Positioning" placement="top-end" open size="lg" trigger="click focus"></sd-tooltip>
       </div>
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="bottom-start" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip
+          content="Bottom Start Positioning"
+          placement="bottom-start"
+          open
+          size="lg"
+          trigger="click focus"
+        ></sd-tooltip>
       </div>
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="bottom" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip content="Bottom Positioning" placement="bottom" open size="lg" trigger="click focus"></sd-tooltip>
       </div>
       <div>
-        <sd-tooltip content="Lorem ipsum" placement="bottom-end" open size="lg" trigger="click focus"></sd-tooltip>
+        <sd-tooltip
+          content="Bottom End Positioning"
+          placement="bottom-end"
+          open
+          size="lg"
+          trigger="click focus"
+        ></sd-tooltip>
       </div>
     </div>
   `
@@ -136,7 +162,7 @@ export const Disabled = {
  * - `manual`: displays the tooltip when the `open` attribute is set to `true`
  */
 export const Trigger = {
-  render: () => html` <sd-tooltip content="Lorem ipsum" placement="bottom" size="lg" trigger="click"></sd-tooltip> `
+  render: () => html` <sd-tooltip content="Lorem ipsum" placement="top-start" size="lg" trigger="click"></sd-tooltip> `
 };
 
 /**
@@ -144,7 +170,7 @@ export const Trigger = {
  */
 export const Hoist = {
   render: () =>
-    html` <div class="flex items-end overflow-scroll h-[15em] w-16">
-      <sd-tooltip content="Lorem ipsum" placement="top" size="lg"></sd-tooltip>
+    html` <div class="flex items-end overflow-scroll h-[5em] w-16">
+      <sd-tooltip content="Lorem ipsum" placement="bottom-start" size="lg"></sd-tooltip>
     </div>`
 };


packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -1,16 +1,18 @@
...
-import { userEvent } from '@storybook/test';
-import { waitUntil } from '@open-wc/testing-helpers';
 import { withActions } from '@storybook/addon-actions/decorator';
 
 const { argTypes, parameters } = storybookDefaults('sd-tooltip');
 const { overrideArgs } = storybookHelpers('sd-tooltip');
 const { generateTemplate } = storybookTemplate('sd-tooltip');
 
+/**
+ * Used to display additional information based on a specific action.
```
MartaPintoTeixeira, line 11:
```suggestion
 * Used as a floating and non-actionable text label that explains an element or function of the user interface.
```

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +67,31 @@ export const Default = {
...
-        {
-          type: 'attribute',
-          name: 'placement',
-          value: 'bottom'
-        }
-      ]
-    })
+  name: 'Size',
+  render: () => html`
+    <div class="flex gap-12">
```
Vahid1919, line 76:
```suggestion
    <div class="flex items-center gap-12">
```

alignment

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +67,31 @@ export const Default = {
...
-          type: 'attribute',
-          name: 'placement',
-          value: 'bottom'
-        }
-      ]
-    });
-  }
+export const Placement = {
+  name: 'Placement',
+  render: () => html` <sd-tooltip content="Lorem ipsum" placement="top" size="lg" trigger="click focus"></sd-tooltip> `
```
Vahid1919, line 88: This only shows the one placement (top) in the story. I would assume all the options would be either in the description, shown here in a flex-box or both.

smfonseca: Placing them all in the story I think would be too much in my opinion, but agree that they can be added in the description.

smfonseca: @Vahid1919 added the text with the options and included a couple of new stories, please review again

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +65,59 @@ export const Default = {
...
-              name: 'placement',
-              value: `<div class="template-height template-position">%TEMPLATE%</div>`
-            }
-          ]
-        })
-      )}
-    </div> `;
-  }
+/**
+ * Use the `size` attribute to set the size.
```
MartaPintoTeixeira, line 69:
```suggestion
 * Use the ”size” attribute to change the tooltip size.
```

smfonseca: Updated

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +65,62 @@ export const Default = {
...
-            type: 'template',
-            name: 'width',
-            value: '<div class="template-position">%TEMPLATE%</div>'
-          }
-        ]
-      })}
-    </div>`;
-  }
+export const Trigger = {
+  render: () => html` <sd-tooltip content="Lorem ipsum" placement="bottom" size="lg"></sd-tooltip> `
```
Vahid1919, line 115:
```suggestion
  render: () => html` <sd-tooltip content="Lorem ipsum" placement="bottom" size="lg" trigger="click"></sd-tooltip> `
```

This might be very minor, but I noticed that the attribute we are describing is not visible under "show code", so I thought it would be more helpful to include it explicitly here even if it is the default.

image

Vahid1919 (approved):

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +65,62 @@ export const Default = {
...
-        { type: 'template', name: 'width', value: '<div class="template-position">%TEMPLATE%</div>' },
-        {
-          type: 'attribute',
-          name: 'placement',
-          value: 'bottom'
-        }
-      ]
-    })
+/**
+ * Use the `placement` attribute to set the placement.
```
MartaPintoTeixeira, line 84: For the placement sample please use the same as in FIGMA

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +65,62 @@ export const Default = {
...
+ * - `bottom-end`: displays tooltip below the trigger element and aligns arrow to the end of the tooltip container
+ */
+export const Placement = {
+  render: () => html`
+    <sd-tooltip content="Lorem ipsum" placement="bottom" size="lg" trigger="click focus"></sd-tooltip>
+  `
 };
 
+/**
+ * Use the `disabled` attribute to disable the tooltip.
```
MartaPintoTeixeira, line 100: Your color in the disabled icon is not matching figma. why is that? do we need to open a fix ticket?
In figma we have Neutral-500

MartaPintoTeixeira, line 100:
```suggestion
 * Use the `disabled` attribute to disable the tooltip trigger.
```

smfonseca: I've checked and `sd-interactive--disabled` is using `neutral-600`. Do we need to update? If so, we need to open an issue

packages/components/src/components/tooltip/tooltip.stories.ts
smfonseca: Updated with a slight change. Left a comment in figma

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +70,81 @@ export const Default = {
...
-          type: 'attribute',
-          name: 'placement',
-          value: 'bottom'
-        }
-      ]
-    })
+/**
+ * Use the `placement` attribute to set the placement.
+ *
+ * - `top-start`: displays tooltip above the trigger element and aligns arrow to the start of the tooltip container
```
MartaPintoTeixeira, line 91: Can you replace Lorem Impsum with the same copy as Figma please?

MartaPintoTeixeira: Also when scrolling on the sample the bottom line of components overlaps the upper one. Can you have a look?

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +70,81 @@ export const Default = {
...
+ * - `top`: displays tooltip above the trigger element
```
MartaPintoTeixeira, line 92: Also when scrolling on the sample the bottom line of components overlaps the upper one. Can you have a look?

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +70,81 @@ export const Default = {
...
-          value: 'bottom'
-        }
-      ]
-    })
+  render: () => html` <sd-tooltip content="Lorem ipsum" placement="top" size="lg" disabled></sd-tooltip> `
 };
 
 /**
- * In case of having the content longer than the space available, the tooltip can be set to scrollable by overriding the `pointer-events` property to auto. In this case the padding-right should also be adjusted via `sd-tooltip::part(body)`. This behavior is heavily discouraged, but it is available in case of need.
+ * Use the `trigger` attribute to control how the tooltip is triggered.
```
MartaPintoTeixeira, line 131: For the tigger sample: can we use a "top-start instead"?
So that when we click the trigger has some margin to the left side.

packages/components/src/components/tooltip/tooltip.stories.ts
```diff
@@ -65,198 +70,81 @@ export const Default = {
...
-  },
-  play: async ({ canvasElement }: { canvasElement: HTMLUnknownElement }) => {
-    const el = canvasElement.querySelector('.mouseless sd-tooltip');
-    await waitUntil(() => el?.shadowRoot?.querySelector('#tooltip'));
-    await waitUntil(() => el?.shadowRoot?.querySelector('sd-icon'));
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    await userEvent.type(el!.shadowRoot!.querySelector('sd-icon')!, '{return}', { pointerEventsCheck: 0 });
-  }
+/**
+ * Use the `hoist` attribute to prevent the tooltip from being clipped when it's placed inside a container with `overflow: auto | hidden | scroll`.
```
MartaPintoTeixeira, line 143: For the hoist sample: can we use a "bottom-start instead"?
So that when we click the trigger has some margin to the left side.

packages/components/src/components/tooltip/tooltip.stories.ts
smfonseca: Updated

packages/components/src/components/tooltip/tooltip.stories.ts
smfonseca: Updated

packages/components/src/components/tooltip/tooltip.stories.ts
smfonseca: Updated

packages/components/src/components/tooltip/tooltip.stories.ts
MartaPintoTeixeira: @smfonseca tag me once the disabled tooltip is Neutral-500. I only need this to approve the ticket..

smfonseca: This will be done once [1364](https://github.com/solid-design-system/solid/issues/1364) is tackled

MartaPintoTeixeira (approved):

PR #1293, opened 2024-08-15
docs: sd-qr-code – add documentation

Vahid1919:
## Description:

Closes ##551

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/docs/Migration/ui-qrcode.mdx
+++ b/packages/components/src/docs/Migration/ui-qrcode.mdx
@@ -1,4 +1,4 @@
-# Migration Guide: Deprecation of `ui-qrcode`
+# Migration Guide: Deprication of `ui-qrcode`
 
 ## [`ui-qrcode`] is deprecated
 
--- a/packages/components/src/docs/Migration/ui-qrcode.mdx
+++ b/packages/components/src/docs/Migration/ui-qrcode.mdx
@@ -2,4 +2,4 @@
 
 ## [`ui-qrcode`] is deprecated
 
-As the ui-qrcode component has seen little use, it will be deprecated and removed from the SDS Library. Should you require a QR code component, we recommend using the [Shoelace QR Code component](https://shoelace.style/components/qr-code) as a replacement.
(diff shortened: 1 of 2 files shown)


packages/components/src/docs/Migration/ui-qrcode.mdx
```diff
@@ -0,0 +1,5 @@
+# Migration Guide: Deprication of `ui-qrcode`
+
+## [`ui-qrcode`] is deprecated
+
+As the ui-qrcode component has seen little use, it will be deprecated and removed from the SDS Library. Should you require a QR code component, we recommend using the [Shoelace QR Code component](https://shoelace.style/components/qr-code) as a replacement.
```
karlbaumhauer, line 5:
```suggestion
As the ui-qrcode component has seen little use, it will be deprecated and removed from the SDS Library. Should you require a QR code component, please open up a feature request and we recommend using the [Shoelace QR Code component](https://shoelace.style/components/qr-code) meanwhile as a replacement.
```

karlbaumhauer: I would like to verify this with @mariohamann as we don't need the suggestions above, if there are technical reasons and also should add them here.

karlbaumhauer (changes requested):

packages/components/src/docs/Migration/ui-qrcode.mdx
```diff
@@ -0,0 +1,5 @@
+# Migration Guide: Deprication of `ui-qrcode`
```
coraliefeil, line 1: # Migration Guide: Deprecation of `ui-qrcode`

coraliefeil:
See my last comment. Approved, when adjusted.

packages/components/src/docs/Migration/ui-qrcode.mdx
mariohamann: The important thign is really more "easy to implement with third-party library or other web component libraries, no benefit from Design side and we want to prevent bigger bundle sizes".

karlbaumhauer (approved):

PR #1322, opened 2024-08-27
docs: sd-input docs [skip chromatic]

auroraVasconcelos:
## Definition of Reviewable:


yoezlem (changes requested):
Please add the "Related Templates" to the component:

image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-        })}
-      </div>
-    `;
-  }
+export const Size = {
+  render: () => html`
+    <div class="grid grid-cols-2 gap-12 content-end">
+      <sd-input size="lg" label="Label"> </sd-input>
+      <sd-input size="md" label="Label"> </sd-input>
+      <sd-input size="sm" label="Label"> </sd-input>
```
yoezlem, line 46: Please fill the placeholders from Figma:
image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-    `;
-  }
+export const Label = {
+  render: () => html`
+    <div class="flex flex-row gap-12 items-baseline">
+      <sd-input label="Label" spellcheck></sd-input>
+      <sd-input spellcheck>
+        <div slot="label" class="text-lg">Label</div>
+      </sd-input>
+    </div>
```
yoezlem, line 64: Please only add input size lg (default) here

image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-            { type: 'attribute', name: 'label', value: 'Password' }
-          ]
-        })}
-      </div>
-    `;
-  }
+export const Disabled = {
+  render: () =>
+    html`<div class="w-[250px]">
+      <sd-input label="Label" placeholder="Input text disabled" disabled></sd-input>
```
yoezlem, line 86: We should also add a "cursor-not-allowed" to the disabled input

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
+    </div>`
 };
 
 /**
- * Use the disabled attribute to disable an input. All interaction is disabled and no events will be fired.
+ * Use the `read-only` attribute to render an input as read only.
```
yoezlem, line 91:
```suggestion
 * Use the `read-only` attribute to render an input as read only. <br> 
 <br>
 Interaction is enabled, but the textarea cannot be edited. Events will be fired.
```

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-      <div class="w-[250px]">
-        ${generateTemplate({
-          constants: [{ type: 'attribute', name: 'readonly', value: true }],
-          args
-        })}
+      <div class="flex flex-rows gap-12">
+        <sd-input label="Label" help-text="Help text"></sd-input>
+        <sd-input label="Label">
+          <div slot="help-text" class="text-lg">Help text</div>
+        </sd-input>
```
yoezlem, line 111: One Input should be enough here (size lg).

image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-        y: { type: 'attribute', name: 'size' }
-      },
-      args
-    });
-  }
+export const Clearable = {
+  render: () =>
+    html`<div class="w-[250px]">
+      <sd-input label="Clearable" clearable spellcheck></sd-input>
+    </div>`
```
yoezlem, line 125: Icon is missing if input is empty. So please fill "Input text" inside the input.

image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
+        ></sd-input>
+      </div>
+      <div class="mb-2">
+        <sd-input
+          type="email"
+          placeholder="someone@example.com"
+          label="Email"
+          help-text="validate with email address format"
+          spellcheck
+        ></sd-input>
```
yoezlem, line 196: Maybe we should change the validation text from german to english?
@mariohamann @coraliefeil 
image

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
-          value: '<sd-icon slot="right" library="global-resources" name="system/picture"></sd-icon>'
-        }
-      ],
-      args
-    });
-  }
+export const Valid = {
+  render: () =>
+    html`<div class="w-[250px]">
+      <sd-input label="Label" style-on-valid></sd-input>
```
yoezlem, line 249: Please fill "Input text here" here, so we can see the valid style out of the box.

image

packages/components/src/components/input/input.stories.ts
auroraVasconcelos: Hey @yoezlem in this case is not a matter of sizes but there are two different examples because the user have two options:

- Using the attibute 
- Using the slot 

I though it would be usefull to have both examples for the code. Do you agree?

packages/components/src/components/input/input.stories.ts
auroraVasconcelos: Same as https://github.com/solid-design-system/solid/pull/1322#discussion_r1738289397

packages/components/src/components/input/input.stories.ts
auroraVasconcelos: @yoezlem from what I understood the language of the error message depends on the browser settings. For example, for me it is in english:
image

packages/components/src/components/input/input.stories.ts
auroraVasconcelos: https://github.com/orgs/solid-design-system/projects/1/views/1?pane=issue&itemId=78382912

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,731 +35,219 @@ export const Default = {
...
+      <sd-input size="lg" label="Label"> </sd-input>
+      <sd-input size="md" label="Label"> </sd-input>
+      <sd-input size="sm" label="Label"> </sd-input>
+    </div>
+  `
 };
 
 /**
- * Add descriptive help text to an input with the `help-text` attribute. For help texts that contain HTML, use the `help-text` slot instead.
+ * Use the `label` attribute to give the input an accessible label.
```
coraliefeil, line 52: Use the "label" attribute to provide an accessible label for the input element.
Use the "label" slot to add custom markup.

packages/components/src/components/input/input.stories.ts
```diff
@@ -69,864 +35,269 @@ export const Default = {
...
+export const Value = {
+  render: () =>
+    html` <div class="w-[250px]">
+      <sd-input value="Value example"></sd-input>
+    </div>`
 };
 
 /**
- * Add the `password-toggle` attribute to add a toggle button that will show the password when activated. Only works with `type="password"`.
+ * Use the `inputmode` attribute Tells the browser what type of data will be entered by the user, allowing it to display the appropriate virtual keyboard on supportive devices.
```
coraliefeil, line 90: * Use the `inputmode` attribute tells the browser what type of data will be entered by the user, allowing it to display the appropriate virtual keyboard on supportive devices.

coraliefeil: The color of the disabled text should be lighter

coraliefeil:
Pattern
Use the "pattern" attribute to use a regular expression to validate the input.

packages/components/src/components/input/input.stories.ts
```diff
@@ -1,932 +1,363 @@
...
-            { type: 'attribute', name: 'label', value: 'Password' }
-          ]
-        })}
-      </div>
-    `;
-  }
+export const Disabled = {
+  render: () =>
+    html`<div class="w-[250px]">
+      <sd-input label="Label" value="Input text disabled" disabled></sd-input>
```
smfonseca, line 110:
```suggestion
      <sd-input label="Label" value="Disabled" disabled></sd-input>
```

smfonseca: To align with what we have been doing in the other components. FYI @MartaPintoTeixeira @coraliefeil

packages/components/src/components/input/input.stories.ts
```diff
@@ -1,932 +1,363 @@
...
-              { type: 'attribute', name: 'placeholder', value: 'https://www.union-investment.de/' },
-              { type: 'attribute', name: 'help-text', value: 'validate with url format' }
-            ],
-            args
-          })}
-        </div>
+export const Type = {
+  render: () =>
+    html` <div class="grid grid-cols-2 gap-12 content-end">
+      <div class="mb-2">
```
smfonseca, line 173: The divs around all sd-selects can be removed

Vahid1919: updated.

coraliefeil:
I’ve updated figma to match code ...

Approved.

packages/components/src/components/input/input.stories.ts
```diff
@@ -1,932 +1,346 @@
...
+ *
+ *  Use the `inputmode` attribute which tells the browser what type of data will be entered by the user, allowing it to display the appropriate virtual keyboard on supportive devices.
+ * - `none`
+ * - `text`
+ * - `decimal`
+ * - `numeric`
+ * - `tel`
+ * - `search`
+ * - `email`
+ * - `url`
```
auroraVasconcelos, line 104: If the story for `inputmode` won't be added please remove the comment about it.

packages/components/src/components/input/input.stories.ts
```diff
@@ -1,932 +1,337 @@
...
+const { argTypes, parameters } = storybookDefaults('sd-input');
 const { generateTemplate } = storybookTemplate('sd-input');
 const { overrideArgs } = storybookHelpers('sd-input');
 
+/**
+ * Used to allow users to enter text. It can be displayed in several ways, depending on the type.
+ *
+ * **Related templates**:
+ * - [Autocomplete](?path=/docs/templates-autocomplete--docs)
+ * - [Input with Currency Stepper](?path=/docs/templates-input-with-currency-stepper--docs)
```
auroraVasconcelos, line 14: For a consistency matter lets just call the template "Input"

packages/components/src/templates/input-with-currency-stepper.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+import type SdInput from '../components/input/input';
+
+/**
+ * ```
+ * ```
+ */
+
+export default {
+  tags: ['!dev'],
+  title: 'Templates/Input with Currency Stepper',
```
auroraVasconcelos, line 12: For a consistency matter lets just call the template "Input"

packages/components/src/templates/input-with-currency-stepper.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+  parameters: {
+    chromatic: { disableSnapshot: true }
+  }
+};
+
+export const Default = {
```
auroraVasconcelos, line 18: Use the "input with currency stepper" title here.

In case we need to add more templates for input in the future we can simply add it here.

PR #1399, opened 2024-09-18
docs: 📚 improve icon docs

mariohamann:
I rewrote the docs to make them work better with our other docs.

### DIFF
--- a/packages/components/src/components/icon/icon.stories.ts
+++ b/packages/components/src/components/icon/icon.stories.ts
@@ -72,7 +72,7 @@ export const Label = {
 };
 
 /**
- * Use the `color` attribute to change an icon’s size
+ * Use the `color` attribute to change an icon’s color
  *
  * - `currentColor` (default): Inherits the current text color, which makes it easy to style icons via CSS.
  * - `primary`: Sets the color to the primary color.
--- a/packages/components/src/components/icon/icon.stories.ts
+++ b/packages/components/src/components/icon/icon.stories.ts
@@ -1,5 +1,5 @@
 import '../../solid-components';
-import { icons } from './library.system';
+import { html } from 'lit-html';
 import { storybookDefaults, storybookHelpers, storybookTemplate } from '../../../scripts/storybook/helper';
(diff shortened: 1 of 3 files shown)


packages/components/src/components/icon/icon.stories.ts
```diff
@@ -8,26 +8,123 @@ const { overrideArgs } = storybookHelpers('sd-icon');
...
+export const Default = {
+  render: (args: any) => {
+    return generateTemplate({ args });
+  }
+};
+
+/**
+ * Use the `name` attribute to change the icon. Checkout these lists of Union Investment's icons for available names:
+ * - [Content Icons](?path=/story/components-sd-icon-default--library-default-content)
+ * - [System Icons](?path=/story/components-sd-icon-default--library-default-system)
```
karlbaumhauer, line 35: do you see another way for the naming? we show the example to register the GR icons down in the story, where we also use the term `system` as the GR library is called system. 

I see a risk of confusion here as it is quite complex anyways and it will be even more complex as soon as we provide the GR libraries in our DS.

karlbaumhauer (approved):
approved but with one suggestion/question about naming.

packages/components/src/components/icon/icon.stories.ts
karlbaumhauer: actually thats what I meant. like the idea of using `internal` instead.

packages/components/src/components/icon/icon.stories.ts
```diff
@@ -8,26 +8,123 @@ const { overrideArgs } = storybookHelpers('sd-icon');
...
+        <sd-button variant="tertiary" size="sm"
+          ><sd-icon name="union-investment/system/arrow-right" label="Next"></sd-icon
+        ></sd-button>
+      </div>
+    `;
+  }
+};
+
+/**
+ * Use the `color` attribute to change an icon’s size
```
coraliefeil, line 75: * Use the `color` attribute to change an icon’s color

yoezlem (approved):

Issue #1355, opened 2024-09-05
fix: 🤔 sd-display main size variant


MartaPintoTeixeira:
## Current behavior
Default size variant in code is XL
Default size variant in figma is 4xl

## Expected behavior
Default to be 4xl

## Technical Information
Figma: https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=2705-24820&node-type=SECTION&t=6LITtAjEN23AWGNx-0

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #1412, opened 2024-09-19
fix: 🤔 sd-display main size variant

yoezlem:
## Description:
The default size variant in code is now changed to 4xl. 
Closes https://github.com/solid-design-system/solid/issues/1355.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/display/display.stories.ts
+++ b/packages/components/src/styles/display/display.stories.ts
@@ -40,17 +40,17 @@ export const Default = {
 /**
  * Use the `&--size-*` classes for alternative appearances.
 
- * - `sd-display--size-4xl`: 4xl can be used as an alternative
+ * - `sd-display`: 4xl is the default display size
  * - `sd-display--size-3xl`: 3xl can be used as an alternative
- * - xl is the default display size
+ * - `sd-display--size-xl`: xl can be used as an alternative
  */
 
 export const Size = {
   render: () => html`
     <div class="flex flex-col gap-6">
-      <div class="sd-display sd-display--size-4xl">Lorem ipsum</div>
-      <div class="sd-display sd-display--size-3xl">Lorem ipsum</div>
       <div class="sd-display">Lorem ipsum</div>
+      <div class="sd-display sd-display--size-3xl">Lorem ipsum</div>
+      <div class="sd-display sd-display--size-xl">Lorem ipsum</div>
     </div>
   `
 };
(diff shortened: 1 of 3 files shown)


karlbaumhauer:
@yoezlem isn't that a breaking change?

yoezlem:
> @yoezlem isn't that a breaking change?

not sure 🤔

@mariohamann @smfonseca ?

auroraVasconcelos:
> > @yoezlem isn't that a breaking change?
> 
> not sure 🤔
> 
> @mariohamann @smfonseca ?

Hey @yoezlem I believe this is a breaking change considering that if you are using the component with the default size you would need to change the code since the default size is now different. 

(Issue link)[https://github.com/solid-design-system/solid/issues/1355] we opened an issue for this when we start re-writting this component, probably we have duplicated issues now fixed by this PR.

smfonseca:
Yes, this is a breaking change imo. It will affect the visual design of the applications where the component is used.

yoezlem:
Changed main branch to feat/breaking-changes

packages/components/src/styles/display/display.stories.ts
```diff
@@ -40,17 +40,17 @@ export const Default = {
 /**
  * Use the `&--size-*` classes for alternative appearances.
 
- * - `sd-display--size-4xl`: 4xl can be used as an alternative
+ * - `sd-display`: 4xl is the default display size
```
auroraVasconcelos, line 43:
```suggestion
 * - 4xl is the default size
```

packages/components/src/styles/display/display.stories.ts
```diff
@@ -40,17 +40,17 @@ export const Default = {
...
  * - `sd-display--size-3xl`: 3xl can be used as an alternative
```
auroraVasconcelos, line 44:
```suggestion
 * - `sd-display--size-3xl`
```

packages/components/src/styles/display/display.stories.ts
```diff
@@ -40,17 +40,17 @@ export const Default = {
...
- * - xl is the default display size
+ * - `sd-display--size-xl`: xl can be used as an alternative
```
auroraVasconcelos, line 45:
```suggestion
 * - `sd-display--size-xl`
```

auroraVasconcelos (approved):
Just to keep consistency with the rest of docs

packages/components/src/styles/display/display.stories.ts
yoezlem: Did we removed the explanation of different variants/classes etc.?

MartaPintoTeixeira (approved):

Issue #1319, opened 2024-08-27
feat: ✨ add vertical orientation to sd-copyright


Vahid1919:
## Description
In Figma the sd-copyright has a vertical variant. In code this variant is missing in order to align the samples (#1246).
This needs to be added.

Figma link: https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=2113-30804&t=XxhVtJuC8ppkMLxn-4

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

Issue #1309, opened 2024-08-23
fix: 🤔 improve copyright visibility


auroraVasconcelos:
## Current behavior
The existing version of the copyright text is white, having a low contrast with light backgrounds/images.
(https://github.com/solid-design-system/solid/pull/1280#issuecomment-2298637967)

## Expected behavior
It is possible to choose between two types of font colors to improve contrast making it more readable.

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch
)

coraliefeil:
@MartaPintoTeixeira I’m done with comp refinement.

IMO we can publish

MartaPintoTeixeira:
All done and published from design side

coraliefeil:
Design is done ... @MartaPintoTeixeira is there sth  @auroraVasconcelos needs to update on dev side?

coraliefeil:
To do: sync docs with code @MartaPintoTeixeira @auroraVasconcelos 
Is copyright now a full component or still a css style?

MartaPintoTeixeira:
feedback provided via call.

PR #1414, opened 2024-09-19
feat: ✨ add vertical variant to sd-copyright

yoezlem:
## Description:
Created now a vertical variant for sd-copyright and marked horizontal as default.
Closes https://github.com/solid-design-system/solid/issues/1319.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/styles/copyright/copyright.css
+++ b/packages/components/src/styles/copyright/copyright.css
@@ -7,7 +7,7 @@
     text-shadow: 0 1px 3px #515151bf;
   }
 
-  &--variant {
+  &--orientation {
     &-vertical {
       &::after {
         @apply pt-2 pb-0 pr-1 pl-0 w-max;
--- a/packages/components/src/styles/copyright/copyright.css
+++ b/packages/components/src/styles/copyright/copyright.css
@@ -6,4 +6,14 @@
     content: var(--copyright);
     text-shadow: 0 1px 3px #515151bf;
   }
+
+  &--variant {
+    &-vertical {
+      &::after {
+        @apply pl-2;
+        writing-mode: vertical-rl;
+        transform: rotate(180deg);
+      }
+    }
+  }
 }
--- a/packages/components/src/styles/copyright/copyright.css
+++ b/packages/components/src/styles/copyright/copyright.css
@@ -12,6 +12,7 @@
       &::after {
         @apply pt-2 pb-0 pr-1 pl-0 w-max;
         writing-mode: vertical-rl;
+        text-orientation: sideways-right;
         transform: rotate(180deg);
       }
     }
--- a/packages/components/src/styles/copyright/copyright.css
+++ b/packages/components/src/styles/copyright/copyright.css
@@ -10,7 +10,7 @@
   &--variant {
     &-vertical {
       &::after {
-        @apply pl-2;
+        @apply pt-2 pb-0 pr-1 pl-0 w-max;
         writing-mode: vertical-rl;
         transform: rotate(180deg);
       }
--- a/packages/components/src/styles/copyright/copyright.declaration.ts
+++ b/packages/components/src/styles/copyright/copyright.declaration.ts
@@ -9,7 +9,7 @@ export default {
     {
       name: 'sd-copyright--orientation-...',
       options: ['vertical'],
-      description: 'The copyrights vertical orientation.'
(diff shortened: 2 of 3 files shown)


yoezlem:
@coraliefeil @MartaPintoTeixeira we need the new variant also now as attribute in figma.

Vahid1919:
@yoezlem I think we should also include a story for the new attribute.

packages/components/src/styles/copyright/copyright.css
```diff
@@ -6,4 +6,14 @@
     content: var(--copyright);
     text-shadow: 0 1px 3px #515151bf;
   }
+
+  &--variant {
```
smfonseca, line 10: I suggest removing `&--variant`. 
`sd-copyright--vertical` is already very descriptive

yoezlem: Avoid making changes in this area. This syntax is also used in other components, such as the container. If the decision is to remove the &--variant, we should create a new ticket to address that.

yoezlem:
> @yoezlem I think we should also include a story for the new attribute.

@Vahid1919 Added.

packages/components/src/styles/copyright/copyright.css
```diff
@@ -6,4 +6,14 @@
...
+    &-vertical {
+      &::after {
+        @apply pl-2;
+        writing-mode: vertical-rl;
```
mariohamann, line 14: Using writing-mode here feels a bit and leads to unexpected errors:

CleanShot 2024-09-19 at 15 13 27@2x

The copyright sign moves strange. I would vote more for

```css
/* Changes for vertical */
.sd-copyright--vertical::after{
  transform: rotate(270deg);
  transform-origin: top left;
  padding-bottom: 0;
  padding-left: 0;
  padding-top: var(--sd-spacing-2,.5rem);
}
```

etc.

Here's a codepen:

https://codepen.io/mariohamann/pen/dyBxJVN?editors=1110

packages/components/src/styles/copyright/copyright.css
mariohamann: We only have variants for color things. We have booleans on other examples as well (see e. g. the headline, hidden-links and interactive). Therefore I second @smfonseca .

mariohamann (changes requested):

packages/components/src/styles/copyright/copyright.css
yoezlem: for map-marker, we also have variants (cluster, main, place). And that was only a quick look in one of the components 🤔
So i would recommend: 
1. New ticket for removing variant for the whole library if its not color related.

packages/components/src/styles/copyright/copyright.css
yoezlem: Okay, but if you reduce the screen size, the copyright text in your example won't break. With vertical-rl, it does.
Changed some styles, maybe you could check it again

packages/components/src/styles/copyright/copyright.css
mariohamann: But there are more than two options, and therefore some kind of "prefix" is needed. It could be called "shape" for "map-marker", which feels kinda artifical.

But here only `horizontal` and `vertical` can exist. And then we usually take booleans.

mariohamann: If you really want to stick to a prefix, at least it should not be variant, but `orientation` in this case:

CleanShot 2024-09-20 at 06 49 05@2x

Would be fine with that approach.

mariohamann (approved):

coraliefeil:
The © is still displayed with an incorrect rotation. Pls fix.

coraliefeil:
> @coraliefeil @MartaPintoTeixeira we need the new variant also now as attribute in figma.

Will be handled via https://github.com/solid-design-system/solid/issues/1309

yoezlem:
> The © is still displayed with an incorrect rotation. Pls fix.

@coraliefeil fixed it. Please check it again. Thx!

coraliefeil:
Approved.

coraliefeil:
> @coraliefeil @MartaPintoTeixeira we need the new variant also now as attribute in figma.

@yoezlem I added it [to this issue ](https://github.com/solid-design-system/solid/issues/1309)

Vahid1919 (approved):

Issue #1420, opened 2024-09-20
docs: 📚 bring back documentation


mariohamann:
## Description

> We're documenting how a component works and what it does and which features it offers – not just single attributes!

During our docs migration, there were former docs migrated to templates (instead of the definition that only samples are moved to templates), that are definitely needed for proper documentation to understand how to use the component:

- [Media / Description](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-media--docs#description): We need to document how to use descriptions, as this is more or less the only thing what media does and why `inverted` exists.
- [Tab Group / Scrollable](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-tabs--docs#tabs%20scrollable): Scrollable is something designers and developers have to be aware of from the documentation. It's part of how it works and they have to do nothing about it.
- [Tab Group / Contained](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-tabs--docs#tabs%20variants): It's worth bringing this into the docs as well. Container is a default feature and therefore has to be shown in context. Move this to `sd-tab` instead of `sd-tab-group`. It's more or less the same as for `sd-option` – it makes not really sense to document `sd-option` `disabled` without the surrounding group.´

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

Vahid1919:
One the sidenote regarding radio-button template. I believe that was created before we decided to turn all radio/radio-button stories into sd-radio-groups. It was not an intentional choice to add it after the fact. But I see how it is somehow "obsolete" now.

PR #1428, opened 2024-09-24
docs: revert docs migrated to templates [skip chromatic]

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/1420

## Definition of Reviewable:


packages/components/src/components/tab/tab.stories.ts
```diff
@@ -55,9 +55,36 @@ export const Default = {
 
 export const Variant = {
   render: () => html`
-    <div class="flex gap-12">
-      <sd-tab variant="default">Default</sd-tab>
-      <sd-tab variant="container">Container</sd-tab>
+    <div class="grid grid-cols-2 gap-12">
+      <sd-tab-group activation="">
```
smfonseca, line 59:
```suggestion
      <sd-tab-group>
```

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -55,9 +55,36 @@ export const Default = {
...
+        <sd-tab-panel name="tab-2">
+          <div>Provident illo neque vel ex.</div>
+        </sd-tab-panel>
+        <sd-tab slot="nav" panel="tab-3" variant="default">Default 3</sd-tab>
+        <sd-tab-panel name="tab-3">
+          <div>Provident illo neque vel ex. Inventore perspiciatis delectus nisi doloremque soluta inventore.</div>
+        </sd-tab-panel>
+      </sd-tab-group>
+
+      <sd-tab-group activation="">
```
smfonseca, line 74:
```suggestion
      <sd-tab-group>
```

smfonseca (approved):

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -68,9 +95,36 @@ export const Variant = {
 
 export const Active = {
   render: () => html`
-    <div class="flex gap-12">
-      <sd-tab active>Default</sd-tab>
-      <sd-tab variant="container" active>Container</sd-tab>
+    <div class="grid grid-cols-2 gap-12">
```
karlbaumhauer, line 98: can we have have the second tab activated? Now it actually looks exactly the same as the story above and therewith its not directly clear to the users, what "active" is actually doing. 
image

karlbaumhauer (changes requested):

packages/components/src/components/tab/tab.stories.ts
auroraVasconcelos: @karlbaumhauer I totally agree with you and that was my first approach but for some reason it just ignores the "active" on second tab and makes the first active

karlbaumhauer: thats strange. Pls have a look into this bug.

auroraVasconcelos: Fixed. There is still an issue but a bugfix ticket for that is created: https://github.com/solid-design-system/solid/issues/1455.

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -91,41 +91,40 @@ export const Variant = {
 
 /**
  * Use the `active` attribute to toggle the active state.
+ *
+ * __Hint:__ Due to an storybook limitation it is necessary if the active tab is the first tab, please refresh the page.
```
karlbaumhauer, line 95:
```suggestion
 * __Hint:__ Due to a Storybook issue, the first tab may be displayed as active. In this case, please reload the page. 
```

karlbaumhauer (approved):
approved besides one small text suggestion.

packages/components/src/components/tab/tab.stories.ts
```diff
@@ -55,23 +55,76 @@ export const Default = {
 
 export const Variant = {
```
MartaPintoTeixeira, line 56: For the variants when tab is open can you use the same copy on both default and container variants?
- Provident illo neque vel ex. Inventore perspiciatis delectus nisi doloremque soluta inventore.
- Lorem ipsum dolor sit amet, consectetur adipiscing elit.
- Inventore perspiciatis delectus nisi doloremque soluta inventore.

MartaPintoTeixeira: Lets also apply this order to the rest of examples to reinforce that tabs as changing please

auroraVasconcelos: Updated

MartaPintoTeixeira (approved):

Issue #1451, opened 2024-09-30
docs: 📚 update supernumber template


auroraVasconcelos:
## Description
Update [storybook](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-supernumber--docs) according to [figma template](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=3724-4897&t=nUqqPqJcrcmXxv4R-4).

## DoR
- [x] Item has business value
- [ ] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #1456, opened 2024-10-01
docs: update supernumber template [skip chromatic]

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/1451.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/templates/supernumber.stories.ts
+++ b/packages/components/src/templates/supernumber.stories.ts
@@ -1,6 +1,10 @@
 import '../solid-components';
 import { html } from 'lit-html';
 
+/**
+ * ```
+ * ```
+ */
 export default {
   tags: ['!dev'],
   title: 'Templates/Supernumber',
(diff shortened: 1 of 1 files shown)


packages/components/src/templates/supernumber.stories.ts
```diff
@@ -1,6 +1,10 @@
 import '../solid-components';
 import { html } from 'lit-html';
 
+/**
```
karlbaumhauer, line 4: was this intentional? If yes, what is it used for?

karlbaumhauer (changes requested):

packages/components/src/templates/supernumber.stories.ts
Vahid1919: @karlbaumhauer Storybook duplicates the default story and we didn't want that behaviour. We have a solution that allows us to hide all default stories in a page, and this is a workaround that prevents the one we want from becoming invisible too.

karlbaumhauer (approved):

MartaPintoTeixeira (approved):

Issue #1452, opened 2024-09-30
docs: 📚 update quote template


auroraVasconcelos:
## Description
Update [storybook](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-quote--docs) according to [figma template](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=4131-17720&t=nUqqPqJcrcmXxv4R-4).

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #1457, opened 2024-10-01
docs: update quote template [skip chromatic]

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/1452.

## Definition of Reviewable:


packages/components/src/templates/quote.stories.ts
```diff
@@ -17,174 +17,103 @@ export default {
...
-    class="w-16 h-16 rounded-full overflow-hidden object-cover ${hidden ? 'hidden' : ''}"
-    src="./placeholders/images/family.jpg"
-    alt="A caring father, balancing one child on his shoulder and gently cradling another in his arms."
-  />
-`;
-
-const information = (hidden = false, text = '4xl') => (hidden ? '' : html`<p class="mt-4 text-sm">${text}</p>`);
-
-/**
- * The default version of a `quote` pattern showcases a quote in an appealing layout,
```
karlbaumhauer, line 31: why did we remove all the information on the quote? as we don't have a component but only the template, I see a need for some further information on how to use it.

packages/components/src/templates/quote.stories.ts
```diff
@@ -17,174 +17,103 @@ export default {
...
-        <div>
-          <p class="sd-paragraph font-bold mb-1">Jane Miller</p>
-          <p class="sd-paragraph sd-paragraph--size-sm">Job Title</p>
-        </div>
-      </div>
-    </div>`
-};
-
-/**
- * This version of the quote provides flexibility in terms of `size` and offers three options: 4xl, 3xl, and xl.
```
karlbaumhauer, line 53: why did we remove all the information on the quote? as we don't have a component but only the template, I see a need for some further information on how to use it.

packages/components/src/templates/quote.stories.ts
```diff
@@ -17,174 +17,103 @@ export default {
...
           </div>
         </div>
-        ${information(false, 'xl')}
-      </div>
-    </div>`
+      </div>`
 };
 
-/**
- * This version, called 'inverted', maintains the flexibility of the quote pattern. It primarily
```
karlbaumhauer, line 103: why did we remove all the information on the quote? as we don't have a component but only the template, I see a need for some further information on how to use it.

packages/components/src/templates/quote.stories.ts
```diff
@@ -17,174 +17,103 @@ export default {
...
+            <p class="sd-paragraph sd-paragraph--size-sm sd-paragraph--inverted">Chairman of the Board</p>
           </div>
         </div>
-        ${information(false, 'xl')}
       </div>
     </div>`
 };
 
-/**
- * The `variant` version of this quote provides additional flexibility by allowing you to toggle
```
karlbaumhauer, line 154: why did we remove all the information on the quote? as we don't have a component but only the template, I see a need for some further information on how to use it.

karlbaumhauer (changes requested):

packages/components/src/templates/quote.stories.ts
MartaPintoTeixeira: Hi Karl, for the old "patterns" we agreed that they would be part of the "templates" but because they are not "components" we created a different structure for them.

packages/components/src/templates/quote.stories.ts
auroraVasconcelos: The main goal is to make it a template, not a component. Quote is an example of how to use the components from the library, not a component.

karlbaumhauer (approved):
@auroraVasconcelos approved as discussed. Pls add the information on the reasoning.

auroraVasconcelos (commented):
As discussed with the team. No comments will be added for now. However, relevant components or styles may be added with the task: https://github.com/solid-design-system/solid/issues/1458.
@karlbaumhauer

MartaPintoTeixeira:
On the menu of templates instead of:
- "Teaser with Icon" lets name it "Teaser"
- "Teaser with Link" lets name it "Teaser"

Both of the samples you have should be in 1 single template the "Teaser"

MartaPintoTeixeira:
Copy change: Instead of "Inverted" lets have "Quote Inverted" please

MartaPintoTeixeira (approved):

Issue #1448, opened 2024-09-30
docs: 📚 update header template


auroraVasconcelos:
## Description
Missing section ["Header Sample A-02"](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=3502-218572&t=nUqqPqJcrcmXxv4R-4) in [storybook](https://solid-design-system.fe.union-investment.de/x.x.x/storybook/?path=/docs/templates-header--docs).

## DoR
- [x] Item has business value
- [ ] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

MartaPintoTeixeira:
The 3rd sample has different copy from FIGMA. But let me check with Coralie if we should actually translate all to english before you edit the copy here.

MartaPintoTeixeira:
When the drawer is open in the hamburger format there is a bug in the last items. Will create a fix ticket to fix the alignment of this bug.

 ![Image](https://github.com/user-attachments/assets/151244fd-13ae-436c-8c73-40b741496d55)

PR #1462, opened 2024-10-08
docs: add sample A - 02 to header template docs [skip chromatic]

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/1448

## Definition of Reviewable:


MartaPintoTeixeira:
After the sample titles can we have the respective description
ex: 1920px – sample/header/lg – variant A - 01

MartaPintoTeixeira:
The 3rd sample has different copy from FIGMA. But let me check with Coralie if we should actually translate all to english before you edit the copy here.

MartaPintoTeixeira:
@auroraVasconcelos I have updated the copy from DE to EN. Please change it accordingly.
https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=3502-217834&node-type=frame&t=LlHtXqbLeW5v1vXG-0

auroraVasconcelos:
> @auroraVasconcelos I have updated the copy from DE to EN. Please change it accordingly. https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=3502-217834&node-type=frame&t=LlHtXqbLeW5v1vXG-0

Updated

MartaPintoTeixeira:
When I adjust screen size (for mobile) i dont see the hamburger icon in this sample.
Can you double check?

MartaPintoTeixeira:
Also the hamburger menu in not working in any of the variants.

packages/components/src/templates/header.stories.ts
```diff
@@ -36,18 +35,19 @@ export const SampleA = {
...
       <div class="hidden lg:flex items-end pt-3 justify-between">
         <!-- bottom-left-area start !-->
         <div class="-ml-4">
-          <sd-navigation-item><b>Über Uns</b></sd-navigation-item
-          ><sd-navigation-item><b>Märkte</b></sd-navigation-item
-          ><sd-navigation-item><b>Presseservice</b></sd-navigation-item>
+          <sd-navigation-item class="font-bold">About Us</sd-navigation-item>
+          <sd-navigation-item class="font-bold">Markets</sd-navigation-item>
+          <sd-navigation-item class="font-bold">Sustainability</sd-navigation-item>
+          <sd-navigation-item class="font-bold">Career</sd-navigation-item>
```
mariohamann, line 50: Please set `href`s here to make this accessible as a link. Apply everywhere.

packages/components/src/templates/header.stories.ts
```diff
@@ -57,41 +57,166 @@ export const SampleA = {
...
+          <sd-navigation-item>
+            <div class="flex items-center gap-2">
+              <sd-icon name="system/lock-locked" class="text-xl"></sd-icon><span>Application</span>
+            </div>
+          </sd-navigation-item>
+          <!-- bottom-right-area end !-->
+        </div>
+      </div>
+    </sd-header>
+    <sd-drawer id="sampleA__drawer" placement="end">
```
mariohamann, line 83: Please write constistent IDs, ideally always in kebap case.

packages/components/src/templates/header.stories.ts
```diff
@@ -57,41 +57,166 @@ export const SampleA = {
               <sd-icon
                 title="Unsere weiteren Auftritte"
                 name="system/website"
-                library="global-resources"
                 class="text-xl absolute -ml-[4px]"
               ></sd-icon>
             </div>
           </sd-navigation-item>
           <sd-navigation-item>
```
mariohamann, line 64: Here we have a little problem: We're proposing a search icon, but do not make any suggestion what should happen, when clicking on it. Should it open a modal? Should it slide out to an text input? Should it do anything at all?

For the moment, I at least would suggest to make an alert, that at least SOMETHING happens. Or you could make it a link, indicating that we would switch to an external search page.

packages/components/src/templates/header.stories.ts
```diff
@@ -57,41 +57,166 @@ export const SampleA = {
...
+    </sd-drawer>
+    <script>
+      // This line is using querySelectorAll due to hidden elements from Storybook configs.
+      // In a regular environment, use: var drawer = document.getElementById('sampleA') or similar
+      var drawerSampleA = document.querySelectorAll('#sampleA__drawer')[1];
+      console.log(document.querySelectorAll('#sampleA__drawer')[1]);
+      function openDrawerSampleA() {
+        drawerSampleA.show();
+      }
+    </script>
```
mariohamann, line 116: Make this a type=module and do not use `var` but const instead.

Remove the console.log.

Why can't you give it a proper ID that is unique? Please adapt, that we can use getElementByID.

Adapt for the other example.

mariohamann (dismissed):

packages/components/src/templates/header.stories.ts
karlbaumhauer: @mariohamann isn't this the case for ALL buttons/items in the header? If think we at least add links to each of the item or (in case of a button) trigger an alert on the page.

packages/components/src/templates/header.stories.ts
auroraVasconcelos: Updated

packages/components/src/templates/header.stories.ts
auroraVasconcelos: href="javascript:void(0)" has been added to all navigation items

packages/components/src/templates/header.stories.ts
auroraVasconcelos: - Var to const updated
- Console.log removed
- type=module is currently breaking, wasn't changed due to it

> Why can't you give it a proper ID that is unique? Please adapt, that we can use getElementByID.
Due to storybook duplication of the stories, there are two ids that are not possible to be prevented. Bug ticket for it: https://github.com/solid-design-system/solid/issues/1566

auroraVasconcelos:
> When I adjust screen size (for mobile) i dont see the hamburger icon in this sample. Can you double check?



> Also the hamburger menu in not working in any of the variants.

Updated

packages/components/src/templates/header.stories.ts
auroraVasconcelos: Hey @karlbaumhauer and @mariohamann,
I believe it's unnecessary to make everything clickable, as this seems to be a template, not a real case example. IMO making every element interactive would add unnecessary complexity to the code, making it harder to clean up and reuse.
For now, I've made only the search icon clickable in each sample. It opens an alert that says, "This could open a search bar.

packages/components/src/templates/header.stories.ts
```diff
@@ -121,38 +245,59 @@ export const SampleB = {
...
+          <sd-icon name="system/website" class="h-6 w-6 mr-2"></sd-icon>
+          Our further appearances
+        </sd-navigation-item>
+      </nav>
+    </sd-drawer>
+    <script>
+      function openDrawerSampleB() {
+        const drawerSampleB = document.getElementById('sample-b-drawer');
+        drawerSampleB.show();
+        0;
```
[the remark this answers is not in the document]
auroraVasconcelos: Updated

MartaPintoTeixeira:
When the drawer is open in the hamburger format there is a bug in the sd-navigation items. Create a fix ticket (https://github.com/solid-design-system/solid/issues/1575) to fix the alignment of this bug.

MartaPintoTeixeira (approved):

karlbaumhauer (dismissed):

MartaPintoTeixeira (approved):
Approved. There is a bug ticket to fix the alignment of the navigation item where there is icon + link (horizontal)

PR #1592, opened 2024-10-25
chore: update failing tests

smfonseca:
## Description:
This PR address two failing tests in the audio component

## Definition of Reviewable:


Vahid1919 (approved):

mariohamann (changes requested):
We try to avoid external files as we this could lead to CDN issues (we already had that.) Isn't there a local way? I would even prefer base64. :D

mariohamann:
Changed title, as no version bump needed.

smfonseca:
> We try to avoid external files as we this could lead to CDN issues (we already had that.) Isn't there a local way? I would even prefer base64. :D

I've added the audio as base64. We also need to update the sd-video tests since it's using an external source. I will create a ticket for that.

mariohamann (approved):
So beautiful ugly. :D

PR #1644, opened 2024-11-14
fix: improve sd-video a11y

paulovareiro29:
## Description:

Closes #1494, #1529

Note: The placeholder video has been replaced by the solid awards video, as suggested by @coraliefeil.

## Definition of Reviewable:

- [x] E2E tests (features, a11y, bug fixes) are created/updated
- [x] Stories (features, a11y) are created/updated
- [x] relevant tickets are linked

PR #1657, opened 2024-11-18
chore!: replace ui-placeholder-video with sds-placeholder-video

paulovareiro29:
## Description:

BREAKING CHANGE: The `ui-placeholder-video.mp4` has been deleted and replaced by `sds-placeholder-video.webm`.

This PR addresses replacing the current placeholder video with the SDS awards video. The decision was made on #1644 

## Definition of Reviewable:


mariohamann:
You should never write the release message in the commit.

paulovareiro29:
> You should never write the release message in the commit, this is handled by Semantic Release in the upcoming release then.

@mariohamann 
Do you mean the `(release/placeholders)`? 

Also, shouldn't the PR name have the `!` in the name? Or does just having `BREAKING CHANGE` on the PR description is enough?

mariohamann:
It has to be either `fix` or `feat`, otherwise it doesn't/shouldn't trigger a release. A `chore` task is not release relevant, as not something that is important for the enduser.

As long as you commit the description a long, it's fine to just have it in the message.

Sidenote: We will change things regarding our releases soon with the switch to Changesets.

paulovareiro29:
As far as I understand from the commitlint docs, as long as there is a `BREAKING CHANGE` mentioned, it will trigger a major release.

[Reference](https://www.conventionalcommits.org/en/v1.0.0/#summary)
`BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.`

paulovareiro29:
UPDATE: An agreement was made with @mariohamann to leave the PR as a chore, even if we are introducing a new image now.
This is to test how the semantic release will behave by using `chore` and `breaking change`.

smfonseca (approved):

PR #1691, opened 2024-11-28
feat: add shadow, position and color variants to copyright

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/1309

## Definition of Reviewable:


paulovareiro29 (approved):

packages/components/src/styles/copyright/copyright.css
```diff
@@ -17,4 +16,27 @@
...
+      &::after {
+        @apply text-black;
+      }
+    }
+  }
+
+  &--shadow {
+    &::after {
+      @apply text-white;
+      text-shadow: 0.5px 0.5px 1.5px #515151;
```
smfonseca, line 31: Looking at Figma I feel this should probably be a drop-shadow. There is a new token `shadow-sm` with the description `Used for copyright text` but it's defined as `box-shadow`so it doesn't work in this case, we would need to update the token and also make some changes in `create-theme`.
@coraliefeil @mariohamann what do you think?

mariohamann: Updating create-theme sounds reasonable.

packages/docs/src/stories/templates/teaser-media.stories.ts
```diff
@@ -142,7 +138,10 @@ export const TeaserMediaWithCopyright = {
       <sd-teaser-media variant="gradient-white" class="max-w-[600px]">
         <img slot="media" src="./placeholders/images/architecture.jpg" class="aspect-video object-cover" alt="" />
         <h3 slot="headline">Headline Media Teaser</h3>
-        <div class="flex flex-col sd-copyright gradient-white" style="--copyright: '© Union Investment 2024'">
+        <div
+          class="flex flex-col sd-copyright sd-copyright--color-black gradient-white"
+          style="--copyright: '© Union Investment 2024'"
```
mariohamann, line 143: This should be wrapped around image like in the examples in the docs?

packages/components/src/styles/copyright/copyright.css
mariohamann, line 1: Please add an appropriate changeset by running `pnpm changeset` .

packages/docs/src/stories/templates/teaser-media.stories.ts
auroraVasconcelos: To be honest I'm not sure what is the best approach. From what I can see this is the easier and cleaner way. If wrapping the image with a div, wich is a slot, or adding the copyright to the teaser media (as the image wrapper), additional CSS is required to fix the styles.

packages/components/src/styles/copyright/copyright.css
auroraVasconcelos: Updated

packages/docs/src/stories/templates/teaser-media.stories.ts
mariohamann: I see – the issue is coming from the fact, that the image is in the background and that the copyright needs it's own space, right?

I better understand your approach now. Looks valid, let's go for it.

packages/components/src/styles/copyright/copyright.css
auroraVasconcelos: Updated

MartaPintoTeixeira (approved):

smfonseca (dismissed):

smfonseca (approved):

Issue #1472, opened 2024-10-09
feat[dev]: ✨ implement A11y improvements to sd-checkbox


yoezlem:
## Description
This task describes the Storybook implementation for the sd-checkbox component as part of the A11y improvements outlined in Epic https://github.com/solid-design-system/solid/issues/1465.

A11y improvements for sd-checkbox can be found here: [sd-checkbox Notes](https://www.notion.so/matuzo/sd-checkbox-107f3f6e1bc080beaa1afefc2a1a55fc).

------------------------------------

## Comment
**Problem:** Invalid checkboxes don't indicate that they're invalid.
**Solution:** Add and toggle aria-invalid="true" on the input element.

**Problem:** Error messages are semantically not connected to the respective input element and are only announced when the user unticks a required checkbox. 
**Solution:** Instead of using a live region for the error message, connect it with the affected input field. Screen readers then announce the message automatically when the user focuses the checkbox. ( **addressed on https://github.com/solid-design-system/solid/pull/1619** )

```
<sd-checkbox size="lg" aria-invalid="true">
  Default Slot
  
  <label part="base">
    <input type="checkbox" required aria-invalid="true" aria-describedby="invalid-message">
    …
  </label>
  
  <div id="invalid-message" part="invalid-message">
    Please tick this box if you want to proceed.
  </div>
</sd-checkbox>
```

## Notes
- Disabled checkboxes: See [@sd-button](https://www.notion.so/sd-button-104f3f6e1bc0800b9f72d2a046316ec3?pvs=24). Everything I wrote there about disabling form elements applies here as well.

## Best Practices
- Don’t nest interactive elements in labels, e.g. links.

## Related WCAG criteria
- https://www.w3.org/WAI/WCAG22/quickref/#error-identification

### Open Questions
- [ ] Question1
- [ ] Question2

## DoR
- [ ] Item has been estimated by the team
- [x] Item dependencies have been identified and documented

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #1707, opened 2024-12-05
fix: sd-checkbox accessible validation state

smfonseca:
## Description:
Closes [1472](https://github.com/solid-design-system/solid/issues/1472)

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/checkbox/checkbox.test.ts
+++ b/packages/components/src/components/checkbox/checkbox.test.ts
@@ -161,13 +161,13 @@ describe('<sd-checkbox>', () => {
       expect(checkbox.hasAttribute('data-valid')).to.be.false;
       expect(checkbox.hasAttribute('data-user-invalid')).to.be.false;
       expect(checkbox.hasAttribute('data-user-valid')).to.be.false;
-      expect(checkbox.getAttribute('aria-invalid')).to.equal('true');
 
       checkbox.click();
       await checkbox.updateComplete;
 
       expect(checkbox.hasAttribute('data-user-invalid')).to.be.true;
       expect(checkbox.hasAttribute('data-user-valid')).to.be.false;
+      expect(checkbox.getAttribute('aria-invalid')).to.equal('true');
     });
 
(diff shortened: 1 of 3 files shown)


packages/components/src/components/checkbox/checkbox.test.ts
```diff
@@ -320,4 +320,22 @@ describe('<sd-checkbox>', () => {
...
+  describe('should contain aria-invalid', () => {
+    it('is false when required and unchecked', async () => {
+      const el = await fixture<HTMLFormElement>(html` <sd-checkbox required></sd-checkbox> `);
+      const input = el.shadowRoot!.querySelector('input')!;
+      expect(input.getAttribute('aria-invalid')).to.equal('false');
+    });
+
+    it('is true when required and checked', async () => {
+      const el = await fixture<HTMLFormElement>(html`<sd-checkbox required checked></sd-checkbox>`);
+      const input = el.shadowRoot!.querySelector('input')!;
```
paulovareiro29, line 333: Question: Shouldn't it be a valid input if it is required and checked? 🤔

packages/components/src/components/checkbox/checkbox.test.ts
paulovareiro29, line 1: Since `aria-invalid` should only when the input is `user-invalid`, maybe we should consider adding tests to test interaction with the checkbox.

Something like what was done on [`textarea`](https://github.com/solid-design-system/solid/blob/4560278baa2eac2bb628ca555b1f724ef94acabe/packages/components/src/components/textarea/textarea.test.ts#L188)

paulovareiro29 (changes requested):

packages/components/src/components/checkbox/checkbox.test.ts
smfonseca: Thanks for catching this

packages/components/src/components/checkbox/checkbox.test.ts
smfonseca: Definitely. Updated accordingly

paulovareiro29 (approved):

Issue #1708, opened 2024-12-06
fix: 🤔 implement `sd-select` support for TalkBack


paulovareiro29:
## Current behavior
When using TalkBack on mobile Android devices, the `sd-select` is not accessible. The user is unable to open the options list.

## Expected behavior
The `sd-select` is fully accessible using TalkBack on Android devices.

## Steps to reproduce
- With an android device, turn on TalkBack.
- Go to the [`sd-select story`](https://solid-design-system.github.io/solid/main/iframe.html?args=&globals=&id=components-sd-select-screenshots-sd-select--default&viewMode=story)
- Try to interact with the `sd-select`

### Suggested Reproduction Tools

- **sd-select story**: https://solid-design-system.github.io/solid/main/iframe.html?args=&globals=&id=components-sd-select-screenshots-sd-select--default&viewMode=story

## Technical Information
From testing multiple design systems, only the [Listbox from Headless UI](https://headlessui.com/v1/vue/listbox) is fully accessible using TalkBack.
I have noticed they use a `<button />` as a trigger for the popup. We are currently using div / input.
We might need to restructure our `sd-select` component to use a `<button />` as a trigger instead, and hide the input just to keep support when using the component within forms

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #1710, opened 2024-12-06
fix: improve sd select a11y [skip chromatic]

### LINKED ISSUE #1486: feat[dev]: ✨ implement A11y improvements to sd-select and sd-option
## Description
This task describes the Storybook implementation for the sd-select and sd-option component as part of the A11y improvements outlined in Epic https://github.com/solid-design-system/solid/issues/1465.

A11y improvements for sd-select and sd-option can be found here: [sd-select and sd-option Notes](https://www.notion.so/matuzo/sd-select-sd-option-10df3f6e1bc080b08bc9dea0e6d78539).

---------------------

## Comment

- ~~**Problem 1:** Building custom selects is rarely a good idea. You took care of many things, but one of the things that makes selects so complex is the variety of shortcuts and how they differ across browsers and screen readers.
**Solution:** Consider switching to [customizable selects ](https://developer.chrome.com/blog/rfc-customizable-select)once they are ready.~~

- **Problem 2:** Invalid fields don't semantically indicate that they're invalid.
**Solution:** Put `aria-invalid`="true" on invalid fields.

- **Problem 3:** Tags: Removing options can be confusing. Mouse users can remove tags by clicking the icon inside the button. Keyboard users can only open the list. That's not perfect, but okay. When you use a screen reader, it gets confusing because the button's label is "Remove option x." However, the button doesn't remove the option; it opens the list. That's what happens in VoiceOver or when you use the Tab key to navigate with other screen readers. When you use the virtual cursor in NVDA, you get "button, option x", and when you move on, you get "button, remove, image, remove". You can activate both, and they'll do what you expect, Again, only when you use the virtual cursor. With the Tab key, you get "remove option x". In JAWS you get “option x remove button” either way.
**Solution:** Split the button and the icon and put the icon in its own button that you position above the other button.
> [!TIP]
> In the combobox, @DanielHargesheimer already implemented a solution. We will follow the same strategy – if he didn't already handle it in his PR.
Test Daniel solution with the screenreader & improve both components if needed.

- **Problem 4:** There is no feedback for screen readers when you remove a tag.
**Solution:** Use a live region to confirm the deletion.

- **Problem 5:** listbox/combobox roles are not supported in Talkback on Android. On top of that, the screen reader announces the combobox as “disabled” because of the readonly attribute.
**Solution 1:** Feature detection (if possible) with a fallback to a native select?
~~**Solution 2:** Use native selects.~~
> [!CAUTION]
> This could be very hard to implement. At least we should investigate some time, to better judge maybe in a follow-up ticket how we proceed there. If this won't be handled in the ticket, we should a disclaimer on the docs page of `sd-select` which informs about and links to the issue.

- **Problem 6:** Once a listbox is open, it is impossible to close it easily using a screen reader.
**Solution:** Allow the closing of lists by pressing Escape.
> [!TIP]
> When we re-tested this in the pre-refinement, it worked. Please have a second look before working on this.
- ~~**Problem 7:** The “Clear entry”-button is not accessible via the keyboard. That’s probably on purpose, but I don’t see why it shouldn’t be focusable.~~
~~**Solution:** Remove the `tabindex`-attribute.~~
> [!NOTE]  
> This is on purpose and therefore shouldn't be touched.

## Notes
Disabled selects/options: See [@sd-button](https://www.notion.so/sd-button-104f3f6e1bc0800b9f72d2a046316ec3?pvs=24). Everything I wrote there about disabling form elements applies here as well.

## Best Practices
- Prefer radio buttons over selects when you only have a few options. Showing is better than hiding.

## Related WCAG criteria
- https://www.w3.org/WAI/WCAG22/quickref/#status-messages
- https://www.w3.org/WAI/WCAG22/quickref/#name-role-value
- https://www.w3.org/WAI/WCAG22/quickref/#labels-or-instructions

## Additional resources
- https://developer.chrome.com/blog/rfc-customizable-select?hl=de

### Open Questions
- [ ] Question1
- [ ] Question2

## DoR

## DoD

paulovareiro29:
## Description:

Closes #1486

This PR addresses:
- Add `aria-invalid` attribute
- Screen reader announcement when a tag is removed

To solve TalkBack issue a new issue has been created: #1708

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/select/select.ts
+++ b/packages/components/src/components/select/select.ts
@@ -218,7 +218,6 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
         @sd-remove=${(event: CustomEvent) => this.handleTagRemove(event, option)}
       >
         ${option.getTextLabel()}
-        <sd-icon library="system" name="close" slot="removable-indicator"></sd-icon>
       </sd-tag>
     `;
   };
--- a/packages/components/src/components/select/select.ts
+++ b/packages/components/src/components/select/select.ts
@@ -97,6 +97,8 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
    */
   @state() private displayLabel = '';
   /** @internal */
+  @state() private liveLabel = '';
+  /** @internal */
   @state() hasFocus = false;
   /** @internal */
   @state() currentOption: SdOption;
@@ -216,6 +218,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
         @sd-remove=${(event: CustomEvent) => this.handleTagRemove(event, option)}
       >
         ${option.getTextLabel()}
+        <sd-icon library="system" name="close" slot="removable-indicator"></sd-icon>
       </sd-tag>
     `;
   };
@@ -510,6 +513,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
 
     if (!this.disabled) {
       this.toggleOptionSelection(option, false);
+      this.liveLabel = this.localize.term('removed', option.textContent);
 
       // Emit after updating
       this.updateComplete.then(() => {
@@ -829,6 +833,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
           this.open && 'z-50'
         )}
       >
+        <span class="sr-only" role="alert" aria-live="assertive">${this.liveLabel}</span>
         <div class="flex items-center gap-1 mb-2">
           <label
             id="label"
@@ -924,6 +929,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
                 aria-haspopup="listbox"
                 aria-labelledby="label"
                 aria-disabled=${this.disabled ? 'true' : 'false'}
+                aria-invalid=${this.showInvalidStyle ? 'true' : 'false'}
                 aria-describedby="help-text invalid-message"
                 role="combobox"
                 tabindex="0"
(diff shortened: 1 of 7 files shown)


packages/components/src/components/select/select.ts
```diff
@@ -97,6 +97,8 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
    */
   @state() private displayLabel = '';
   /** @internal */
+  @state() private liveLabel = '';
```
smfonseca, line 100: I feel it would be better to give it a more descriptive name, something like `deletedTagLabel`.

packages/components/src/components/select/select.ts
```diff
@@ -924,6 +929,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
                 aria-haspopup="listbox"
                 aria-labelledby="label"
                 aria-disabled=${this.disabled ? 'true' : 'false'}
+                aria-invalid=${this.showInvalidStyle ? 'true' : 'false'}
```
smfonseca, line 932:
```suggestion
                aria-invalid=${this.showInvalidStyle}
```

packages/components/src/translations/de.ts
```diff
@@ -38,7 +38,8 @@ const translation: Translation = {
   showPassword: 'Passwort anzeigen',
   slideNum: (slide, count) => `Folie ${slide} von ${count}`,
   toggleColorFormat: 'Farbformat umschalten',
-  unmute: 'Unmute'
+  unmute: 'Unmute',
+  removed: name => `${name} entfernt`
```
smfonseca, line 42: Please sort it alphabetically. Same in de `de` and `en` files.

smfonseca (approved):

PR #1712, opened 2024-12-06
fix: make sd-tooltip more accessible

### LINKED ISSUE #1493: feat[dev]: ✨ implement A11y improvements to sd-tooltip
## Description
This task describes the Storybook implementation for the sd-tooltip component as part of the A11y improvements outlined in Epic https://github.com/solid-design-system/solid/issues/1465.

A11y improvements for sd-tooltip can be found here: [sd-tooltip Notes](https://www.notion.so/matuzo/sd-tooltip-10ef3f6e1bc080b5a2f2c4d4a6076500).

------------------

## Comment
**Problem:** Users must click the button twice the first time to toggle it.
  - [x]  **Solution:** Toggle with the first click.

**Problem:** The button doesn't have an accessible name.
  - [x]  **Solution:** Provide a label for the icon.

**Problem:** By definition, a tooltip is "A contextual popup that displays a description for an element.", but the button in this component is not connected to the tooltip.
  - [x]  **Solution:** Put `aria-describedby="tooltip"` on the button, not the slot.
-> We have to add a new "story" in docs (Storybook & Figma) which shows tooltip being overwritten with the default button as a trigger and then showing how to connect button by `aria-describedby="tooltip"`. Some limitations are left here: https://github.com/solid-design-system/solid/issues/1700

**Problem:** Using the virtual cursor, screen reader users access the tooltip content before the button in NVDA and VO. It should be the other way around.
  - [ ]  **Solution:** I'm not exactly sure what's causing this problem (and probably others), but I think it's because you're sloting a slot element.

`<slot slot="anchor"></slot>`
-> wrap a div around:
`<div slot="anchor"><slot>Hello World</slot></div>`

**Problem:** Disabled buttons are focusable and operable (although in an unexpected way; see video).
  - [x]  **Solution:**  Disable them properly.
Video Example: https://www.notion.so/matuzo/sd-tooltip-10ef3f6e1bc080b5a2f2c4d4a6076500?pvs=4#10ef3f6e1bc080bf9518ea23fdb012d9

Problem: Scrollable content is not accessible to keyboard users because the tooltip closes when you try to tab into it to scroll.
  - [x]  Solution:  Make the scrollable area keyboard focusable and don't close the tooltip when focus moves.
-> make it only focusable if content is overflowing.
  **Recommendation to only include short text is added to the Do's and Don'ts.**

## Best Practices
- Don’t use tooltips for critical or unique information.
- Avoid long text in tooltips.

## Related WCAG criteria
- https://www.w3.org/WAI/WCAG22/quickref/#content-on-hover-or-focus
- https://www.w3.org/WAI/WCAG22/quickref/#keyboard
- https://www.w3.org/WAI/WCAG22/quickref/#name-role-value

## Additional resources
- https://dequeuniversity.com/library/aria/tooltip

### Open Questions
- [ ] Question1
- [ ] Question2

## DoR

## DoD

Vahid1919:
## Description:

Closes #1493 

Changes include:
- The tooltip is toggled starting the first click, instead of ignoring it.
- Provide a label for the icon.
- Move aria-describedby="tooltip" on the button
- Remove "Long Content" story from screenshot tests

Screen Reader Test Results: 
* VO with Chrome: "Tooltip, Lorem Ipsum, Button, Group, Tooltip" -- both "Tooltip" comes from the label of the icon.  
* VO with Firefox: "Tooltip, button, group, Lorem Ipsum"
* VO with Safari: Screen Reader skips the tooltip entirely. This happened before the changes as well. 
* NVDA on Windows (chrome): "Lorem Ipsum, clickable tooltip graphic button, lorem ipsum" -- Why this happens is a bit unclear, the suggestion Manual gave does not work with slots. 

## Definition of Reviewable:


Vahid1919:
@MartaPintoTeixeira I added a new story for the tooltip docs at the end. Please take a look and let me know if I should make any changes before we align it with Figma :) 
 https://solid-design-system.github.io/solid/fix_sd-tooltip-a11y/iframe.html?viewMode=docs&id=components-sd-tooltip--docs#custom%20buttons

MartaPintoTeixeira:
Custom Buttons section:

Copy suggestion: Note: Currently, this method is not recommended due to issues with passing ID references to the shadow DOM, which can make the tooltip text inaccessible to screen readers. We are actively working on a solution to resolve this limitation.

MartaPintoTeixeira:
Something is wrong in the 1st tooltip sample
Screenshot 2025-01-14 at 15 42 37

MartaPintoTeixeira:
Samples inside the Placement section have a weird behaviour in responsive screens (this is also happening in the current storybook link) should we open a ticket for this?

MartaPintoTeixeira:
As discussed via call:
- in the template instead of "Home page" write "Home" as in figma
- to prevent the accessibility of 1st and last sd-navigation item lets add a padding top of 4 before the 1st group of sd-navigation items and a padding of 4 after the sd-navigation items of the footer section
- left and right paddings to be dicussed in the daily

MartaPintoTeixeira:
Component is still broken. Assign me again once fixed.
Screenshot 2025-01-28 at 14 39 10

MartaPintoTeixeira (changes requested):
Component is still broken. Assign me again once fixed.

smfonseca:
> Component is still broken. Assign me again once fixed.

Fix is in [#1798](https://github.com/solid-design-system/solid/pull/1798)

MartaPintoTeixeira (changes requested):
In the tmplt: Select with Tooltip
can you align it as Top Start Positioning? Because now is not perfectly centered.

smfonseca (commented):
Please double check click after hover as discussed on call

packages/components/src/components/tooltip/tooltip.test.ts
```diff
@@ -184,50 +186,78 @@ describe('<sd-tooltip>', () => {
...
 
-    showSpy.restore();
-    hideSpy.restore();
+    // tooltip is hidden again after clicking the button a second time
+    expect(body.hidden).to.be.true;
+    expect(document.activeElement).to.not.equal(el);
   });
 
-  it('removes whitespace nodes from the default slot', async () => {
+  it('should close tooltip when clicking trigger if it was opened by focus', async () => {
```
smfonseca, line 235: Since the trigger is only `focus`, clicking the button shouldn't hide the tooltip, this is only happening because on L256 you're using `blur()`. I think it would be easier to check the `open` attribute when focusing on the button and then when removing focus.

packages/components/src/components/tooltip/tooltip.test.ts
```diff
@@ -184,50 +186,78 @@ describe('<sd-tooltip>', () => {
...
+    // tooltip starts hidden
+    expect(body.hidden).to.be.true;
 
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
     button.focus();
-    await waitUntil(() => showSpy.calledOnce);
+    await waitUntil(() => !body.hidden);
 
-    // tooltip is visible after focusing
+    // tooltip is visible after clicking the button
```
smfonseca, line 223:
```suggestion
    // tooltip is visible after focusing the button
```

packages/components/src/components/tooltip/tooltip.test.ts
```diff
@@ -184,50 +186,78 @@ describe('<sd-tooltip>', () => {
...
-    expect(hideSpy.notCalled).to.be.true;
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
+    button.blur();
+    await waitUntil(() => body.hidden);
 
-    showSpy.restore();
-    hideSpy.restore();
+    // tooltip is hidden again after clicking the button a second time
+    expect(body.hidden).to.be.true;
+    expect(document.activeElement).to.not.equal(el);
```
smfonseca, line 232: I think this assertion is not needed

mariohamann: Wasn't there a bug that triggered this behaviour?

auroraVasconcelos: @smfonseca @mariohamann,
There was one specific situation we would have this bug:
image (75)

Therefore I added this situation to the tests in order to prevent this bug to happen again

packages/docs/src/stories/components/tooltip.stories.ts
```diff
@@ -112,8 +107,9 @@ export const Size = {
  */
 
 export const Placement = {
+  // parameters: { ...parameters, docs: { story: { inline: false, height: '800px' } } },
```
paulovareiro29, line 110: remove dead code if not necessary

paulovareiro29 (changes requested):
LGTM
Just some commented code that needs to be removed if not necessary

MartaPintoTeixeira (dismissed):

packages/docs/src/stories/components/tooltip.stories.ts
auroraVasconcelos: Updated

paulovareiro29 (approved):

PR #1759, opened 2025-01-07
ci: fix failling verify monorepo

paulovareiro29:
## Definition of Reviewable:

### DIFF
--- a/packages/components/scripts/build.js
+++ b/packages/components/scripts/build.js
@@ -126,7 +126,7 @@ async function nextTask(label, action) {
 }
 
 await nextTask('Cleaning up the previous build', async () => {
-  await Promise.all([...bundleDirectories.map(dir => deleteAsync(dir))]);
+  await Promise.all([...bundleDirectories.map(dir => deleteAsync(dir, { force: true }))]);
   await fs.mkdir(outdir, { recursive: true });
 });
 


smfonseca (dismissed):
@paulovareiro29 please add an empty changeset. Thanks

paulovareiro29:
> If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**

@smfonseca I don't believe we need to add empty changesets when we don't want to version bump, however I have added it.

> If these changes should not result in a new version, you're good to go. If these changes should result in a version bump, you need to add a changeset.

smfonseca:
We don't need to add an empty changeset, but from my perspective, it's more of a best practice that all our PRs have a changeset even if it's empty, that way changeset-bot (which is "kind of" a check) is also passing

packages/components/scripts/build.js
```diff
@@ -12,6 +13,7 @@ import { litTailwindAndMinifyPlugin } from './esbuild-plugin-lit-tailwind-and-mi
 const outdir = 'dist';
 const cdndir = 'cdn';
 const lite = process.argv.includes('--lite');
+const watch = process.argv.includes('--watch');
```
mariohamann, line 16: Do we really need `lite` AND `watch`?

paulovareiro29: I feel like we should separate just building a dist folder (lite) from the watch feature.
If for some reason, watch would be used by anyone and they needed the cdn folders to be generated, it wouldn't.

smfonseca (approved):

PR #1902, opened 2025-02-12
fix: improve sd-notification a11y

### LINKED ISSUE #1482: feat[dev]: ✨ implement A11y improvements to sd-notification
## Description
this task describes the Storybook implementation for the sd-notification component as part of the A11y improvements outlined in Epic https://github.com/solid-design-system/solid/issues/1465.

A11y improvements for sd-notification can be found here: [sd-notification Notes](https://www.notion.so/matuzo/sd-notification-10ef3f6e1bc08069a42fff79906f5254).

- [x] First, we have to solve #1693 
- [x] [See updated Figma docs](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=2265-6175&t=tZ9yNamTo7QVlkrU-11) which contain information of how to build toasts in an accessible way.

**We need to cover both requirements/notes in Figma and this ticket. In case we have conflicting suggestions/infos, we will prefer Figma.**

-----------

## Comment
**Problem:** Announcements are inconsistent, insufficient, and confusing.
- VoiceOver on macOs announces "Close"
- Talkback on Android announces "lorem ispum close".
- NVDA doesn't announce anything except when another message is already present; then it announces "notification, lorem ipsum close button" for the new message.
- JAWS announces, "notification, lorem ipsum close"

Screen reader users either don't receive the message or get a message that includes the button label, which some users could interpret as an action they must take.

One of the reasons for the inconsistency is that with some screen readers, live regions added to the page after the page has loaded are sometimes ignored. Another is that live regions don't announce roles, only text.

**Solution 1:** Use an empty wrapper div with `role=status` that's always present for your messages. Notifications that appear inside the div will be read out. Avoid interactive elements inside your notifications.
**Solution 2:** Don't use toasts.

**Problem:** People who use software to zoom into parts on the screen may miss toast messages.
**Solution 1:** Don't use toasts for critical information. Always show critical notifications inline on the page.
**Solution 2:** Don't use toasts.

**Problem:** When the page is zoomed in, toasts may obscure large parts of the screen.
**Solution 1:** Don't use toasts and show notifications only inline.

**Problem:** Keyboard users must tab through the remaining page to close toasts or use other controls within.
**Solution 1:** Don't move focus, don't use toasts for critical information and actions, and try to avoid using interactive elements in toasts.
**Solution 2:** Don't use toasts.

## Notes
- If someone were to ignore, or miss a toast message due to its timed display, there should be no negative impact on their current activities or the status that the message conveyed. Using the previous examples, ignoring a toast message would still mean that a file was saved, that a message was sent, or that a meeting was about to start. (Scott O’Hara)
- Consider not using buttons in timed toasts. [Scott O’Hara explains why](https://www.scottohara.me/blog/2019/07/08/a-toast-to-a11y-toasts.html#inclusive-ux-of-a-toast).

## Best Practices
- Notification shouldn’t close earlier than 5 seconds.
- Don’t use toasts for critical information.
- Prefer default (inline) notifications over toast notifications.

## Related WCAG criteria
- https://www.w3.org/WAI/WCAG22/quickref/#timing-adjustable
- https://www.w3.org/WAI/WCAG22/quickref/#reflow
- https://www.w3.org/WAI/WCAG22/quickref/#resize-text
- https://www.w3.org/WAI/WCAG22/quickref/#status-messages

## Additional resources
- https://www.scottohara.me/blog/2019/07/08/a-toast-to-a11y-toasts.html
- https://adrianroselli.com/2020/01/defining-toast-messages.html
- https://adamsilver.io/blog/the-problem-with-toast-messages-and-what-to-do-instead/

## DoR

## DoD

paulovareiro29:
## Description:

Closes #1482

## Definition of Reviewable:

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,5 +1,20 @@
 # @solid-design-system/components
 
+## 4.1.1
+
+### Patch Changes
+
+- Make `sd-tooltip` more accessible: _[`#1712`](https://github.com/solid-design-system/solid/pull/1712) [`776c0e4`](https://github.com/solid-design-system/solid/commit/776c0e4579644380b0b169c76373b522ce508edf) [@Vahid1919](https://github.com/Vahid1919)_
+
+  - Fixed tooltip ignoring first click due to focus logic.
+  - Provided a label for the icon.
+  - Added `aria-describedby` attribute to the button.
+  - Removed "Long Content" story from screenshot tests.
+
+### 📈 Stats
+* Uncompressed: 403 KB (+1 KB / +0%)
+* Gzipped: 95 KB (unchanged)
+
 ## 4.1.0
 
 ### Minor Changes
@@ -9,8 +24,9 @@
   The new icons can be seen [here](https://solid-design-system.fe.union-investment.de/docs/?path=/story/components-sd-icon-default--status-library).
 
 ### 📈 Stats
-* Uncompressed: 402 KB (+2 KB / +0%)
-* Gzipped: 95 KB (unchanged)
+
+- Uncompressed: 402 KB (+2 KB / +0%)
+- Gzipped: 95 KB (unchanged)
 
 ## 4.0.10
 
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@solid-design-system/components",
   "description": "Solid Design System: Components",
-  "version": "4.1.0",
+  "version": "4.1.1",
   "homepage": "https://solid-design-system.fe.union-investment.de/docs/",
   "author": "Union Investment",
   "license": "MIT",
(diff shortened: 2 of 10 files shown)


MartaPintoTeixeira:
Question:
On the "Toast Notifications" section
When i click in the 1st samples (that show top-right) the toast notification have a duration.
When i click in the 2nd samples (bottom-center) they stay until I close.

Was this on purpose? 
If you I would remove the duration of the samples above since the purpose of this section is only to show the placement options and the duration is covered before this section

MartaPintoTeixeira (dismissed):
Please have a look at my previous comment and change it if possible.
In any cases consider it approved.

auroraVasconcelos (approved):
LGTM

smfonseca (approved):

MartaPintoTeixeira (approved):

PR #1920, opened 2025-02-24
feat: highlight all search occurences

### LINKED ISSUE #1001: feat[dev]: ✨ highlight ALL instead of 1st occurences of search term in results
## Description
The current `ui-autosuggest` provides a way to highlight all occurences of the search term in the results. The `combobox` template only shows the first occurence. 

So e.g. "un" would be marked twice in "**Un**iD**un**i" => case-insensitive

Image

## DoR
- [x] Item has business value

## DoD

paulovareiro29:
## Description:

Closes #1001

## Definition of Reviewable:


mariohamann:
From my understanding, here is a misunderstanding. I'm not sure, if this should be the default behaviour. In this comment https://github.com/solid-design-system/solid/issues/1001#issuecomment-2494134531 I meant, that we could create a custom renderer in a template here. IMO it shouldn't be the default behaviour. But I'm not sure? @coraliefeil 

Furthermore: Could you please add a ticket that we change this to use the Highlight API? https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API This enables it to not change the DOM here.

coraliefeil:
@mariohamann It’s on dev side to decide if this is default or provided as template.
We agreed on the template I believe.

If I added it in the component by mistake I’ll remove it there.

mariohamann:
Hm. I would see this more as a UX question, or am I wrong here?

What's interesting: I had a look on all those 39 examples here (https://component.gallery/components/combobox/) – and only the following implemented highlighting AT ALL:
- https://boltdesignsystem.com/pattern-lab/?p=viewall-components-typeahead
- https://cauldron.dequelabs.com/components/Combobox
- https://eui.elastic.co/#/forms/combo-box

CleanShot 2025-02-24 at 14 36 43@2x

CleanShot 2025-02-24 at 14 36 21@2x

CleanShot 2025-02-24 at 14 39 20@2x

There was three others, that highlighted ALL occurrences:

- https://designsystem.morningstar.com/components/combo-box/?version=6.0.0&tab=variations&fwk=Vue%202
- https://pharos.jstor.org/components/combobox
- https://reach.tech/combobox/

CleanShot 2025-02-24 at 14 41 31@2x

CleanShot 2025-02-24 at 14 45 25@2x

CleanShot 2025-02-24 at 14 46 07@2x

Personally, for me on ehighlight is totally enough – I don't need multiple occurencies or they even feel to overload the UI. But yah, I would say the default should be a UX decision. :)

MartaPintoTeixeira:
I found a bug. To replicate do as suggested:

1 ) in Multiple Highlight Query type "or" - it will highlight all the "or" which is correct
2 ) now click outside the combobox you will see that the "or" stays in there and it should not.

As a desired result check what happens in the Highlight Query when same steps are take.

MartaPintoTeixeira:
The bug mentioned above is now fixed.

MartaPintoTeixeira (approved):

mariohamann (approved):

Issue #1978, opened 2025-03-12
fix: sd-header


smfonseca:
## Current behavior
- Buttons must have discernible text (button-name)
- Document should not have more than one banner landmark (landmark-no-duplicate-banner)
- Landmarks should have a unique role or role/label/title (i.e. accessible name) combination (landmark-unique)

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #2001, opened 2025-03-13
fix: sd-header a11y

smfonseca:
## Description:
Closes [#1978](https://github.com/solid-design-system/solid/issues/1978)

## Definition of Reviewable:

### DIFF
--- a/packages/docs/src/stories/components/header.test.stories.ts
+++ b/packages/docs/src/stories/components/header.test.stories.ts
@@ -15,7 +15,7 @@ const { generateScreenshotStory } = storybookUtilities;
 
 export default {
   title: 'Components/sd-header/Screenshots: sd-header',
-  tags: ['!autodocs', 'skip-a11y-[landmark-no-duplicate-banner]', 'skip-a11y-[landmark-unique]'],
+  tags: ['!autodocs', 'skip-a11y-[landmark-no-duplicate-banner, landmark-unique]'],
   component: 'sd-header',
   args: overrideArgs([
     {
--- a/packages/docs/src/stories/components/header.test.stories.ts
+++ b/packages/docs/src/stories/components/header.test.stories.ts
@@ -186,10 +186,10 @@ export const SampleA = {
             <!-- bottom-right-area start !-->
             <div class="-mr-4 flex items-center">
               <sd-navigation-item>
-                <div class="w-4 h-4 flex items-center"><sd-icon title="Unsere weiteren Auftritte" name="system/website" class="text-xl absolute -ml-[4px]" label="Other locations"></sd-icon></div>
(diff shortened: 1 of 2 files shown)


paulovareiro29 (changes requested):

packages/docs/src/stories/components/header.test.stories.ts
```diff
@@ -163,10 +186,10 @@ export const SampleA = {
             <!-- bottom-right-area start !-->
             <div class="-mr-4 flex items-center">
               <sd-navigation-item>
-                <div class="w-4 h-4 flex items-center"><sd-icon title="Unsere weiteren Auftritte" name="system/website" class="text-xl absolute -ml-[4px]"></sd-icon></div>
+                <div class="w-4 h-4 flex items-center"><sd-icon title="Unsere weiteren Auftritte" name="system/website" class="text-xl absolute -ml-[4px]" label="Other locations"></sd-icon></div>
```
mariohamann, line 189: I believe we can remove the title-attrbute then, right?

smfonseca: Not necessarly. The title provides visual indication for people that can see the icon but don't understand it. I would keep it. What do you think @mariohamann?

mariohamann: Ah. Hm. Do we do this everywhere, where we `sd-icon` without text? I would like to follow a general rule here. Either everywhere or nowhere.

From my feeling `title` is a false promise – it's e. g. not working for Mobile Users and therefore fixes only something for some...

smfonseca: Having a quick look on the codebase, I think we mostly don't use it, so to keep the consistency I will also remove it here.

> not working for Mobile Users and therefore fixes only something for some

This is true but can be said for other things aswell, we unfortunately aren't able to cover everything.

packages/docs/src/stories/components/header.test.stories.ts
```diff
@@ -163,10 +186,10 @@ export const SampleA = {
...
               </sd-navigation-item>
               <sd-navigation-item>
-                <div class="w-4 h-4 flex items-center"><sd-icon title="Suche" name="system/magnifying-glass" class="text-xl absolute -ml-[4px]"></sd-icon></div>
+                <div class="w-4 h-4 flex items-center"><sd-icon title="Suche" name="system/magnifying-glass" class="text-xl absolute -ml-[4px]" label="Search"></sd-icon></div>
```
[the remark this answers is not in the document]
smfonseca: Removed

paulovareiro29 (approved):

PR #2025, opened 2025-03-17
fix: sd-copyright shadow not rendering

smfonseca:
## Description:
This PR fixes the shadow of `sd-copyright` which was not rendering due to missing variables.
The issue is only visible outside of storybook (eg: codepen) and was reported in the support channel.

## Definition of Reviewable:

### DIFF
--- a/packages/styles/src/solid-styles.css
+++ b/packages/styles/src/solid-styles.css
@@ -10,3 +10,14 @@
 @import url('./modules/status-badge.css');
 /* plop:style */
 @import url('./modules/prose.css');
+
+body {
+  --tw-blur: ;
+  --tw-brightness: ;
+  --tw-contrast: ;
+  --tw-grayscale: ;
+  --tw-hue-rotate: ;
+  --tw-invert: ;
+  --tw-saturate: ;
+  --tw-sepia: ;
+}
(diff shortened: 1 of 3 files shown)


packages/styles/src/solid-styles.css
```diff
@@ -10,3 +10,14 @@
...
+body {
+  --tw-blur: ;
+  --tw-brightness: ;
+  --tw-contrast: ;
+  --tw-grayscale: ;
+  --tw-hue-rotate: ;
+  --tw-invert: ;
+  --tw-saturate: ;
+  --tw-sepia: ;
+}
```
mariohamann, line 23: That looks odd. 😆 

I'm irritated, that there's no fallback in the classes or when applying. Can you have more a look there?

And if we have to do it like this, I would maybe go with `:root`. But I believe, this shouldn't be done like this....

mariohamann:
My change seems to work: 
CleanShot 2025-03-20 at 12 10 35@2x

smfonseca:
> My change seems to work

Yes, but shouldn't this be global? Because we will have the same problem if we need these variables in other components. And the increase in bundle size will be minimal from my point of view

smfonseca:
> > My change seems to work
> 
> Yes, but shouldn't this be global? Because we will have the same problem if we need these variables in other components. And the increase in bundle size will be minimal from my point of view

Discussed on teams call. Will keep like this and create a ticket to improve screenshot tests to catch this kind of bugs.

PR #2086, opened 2025-03-26
fix: sd-select refactor

smfonseca:
## Description:
Closes [#1708](https://github.com/solid-design-system/solid/issues/1708)

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/select/select.ts
+++ b/packages/components/src/components/select/select.ts
@@ -515,7 +515,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
       }
 
       // Set focus after updating so the value is announced by screen readers
-      this.updateComplete.then(() => this.displayInput.focus({ preventScroll: true }));
+      this.updateComplete.then(() => this.combobox.focus({ preventScroll: true }));
 
       if (this.value !== oldValue) {
         // Emit after updating
@@ -527,7 +527,7 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
 
       if (!this.multiple) {
         this.hide();
-        this.displayInput.focus({ preventScroll: true });
+        this.combobox.focus({ preventScroll: true });
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/components/src/components/select/select.ts
```diff
@@ -980,131 +984,144 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
...
-              @keydown=${this.handleComboboxKeyDown}
-              @mousedown=${this.handleComboboxMouseDown}
-              @mouseenter=${this.handleMouseEnter}
-              @mouseleave=${this.handleMouseLeave}
-            >
-              <input
-                name=${this.name}
-                form=${this.form}
-                part="display-input"
+            <div class="relative w-full h-full grid grid-cols-1" slot="anchor">
```
mariohamann, line 987: Wait. So we can now use different triggers as well? Is this useful?

packages/components/src/components/select/select.ts
```diff
@@ -223,6 +224,8 @@ export default class SdSelect extends SolidElement implements SolidFormControl {
         removable
         @keydown=${(event: KeyboardEvent) => this.handleTagKeyDown(event, option)}
         @sd-remove=${(event: CustomEvent) => this.handleTagRemove(event, option)}
+        @mouseenter=${this.handleMouseEnter}
```
mariohamann, line 227: I just realized, that the select feels really laggy. I saw, that the hover styles are triggered by JavaScript, instead of `hover:`. Could we refactor that in this PR? (Maybe ahs to be changed in Combobox as well.)

smfonseca: Updated both `sd-select` and `sd-combobox`.
Hover was being applied with js because of some issue in Safari but I could not replicate so the switch should be ok.

I think the component is laggy in storybook because it has many stories. I tested by leaving just the default and it felt much better.

packages/components/src/components/select/select.ts
smfonseca: No, that is not the case, we already had it setup similar to this before

mariohamann (approved):

Issue #2054, opened 2025-03-21
feat[dev]: ✨ implement A11y Re-Improvements to sd-input


yoezlem:
## Description
This tasks describes the Figma implementation for: #2048 

This task describes the Storybook implementation for the sd-input component as part of the A11y re-improvements outlined in Epic #2048.

A11y improvements for sd-input can be found here: [sd-input Notes](https://www.notion.so/matuzo/sd-input-1baf3f6e1bc08100b262d88e6a242c24).

------------------------------------------

**Problem:** Invalid fields don't semantically indicate that they're invalid.
**Solution:** Put `aria-invalid="true"` on invalid fields.

**Problem:** The “Show password”-button is not accessible via the keyboard.
**Solution:** Remove the `tabindex`-attribute.

## Notes
- Instead of (or in addition to) changing the label, you could also toggle the `aria-pressed` attribute to indicate whether the password is visible.
- Please note that `maxlength` doesn't indicate the maximum number of characters to screen readers. If there's no help text, it can be confusing.
- Consider noting in the docs that the placeholder is not a replacement for the label but only an addition.
- Disabled inputs: See [[sd-button](https://www.notion.so/sd-button-1baf3f6e1bc08198a2d6ddf4df8b1e21?pvs=21)](https://www.notion.so/sd-button-1baf3f6e1bc08198a2d6ddf4df8b1e21?pvs=21) . Everything I wrote there about disabling form elements applies here as well.

## Accessibility information
- *Ensure that focus moves logically between input fields and other form elements. When an input field is focused, it should be clearly indicated.*
**”logically” can be misleading. Consider using “in sequential order” instead.**
- *If an input field contains an error, ensure that it is announced by screen readers. Use aria-invalid="true" and aria-describedby to link the input field to the error message.*
**Isn’t that something the component already does for me?**
- Use the right type for the given purpose (email, url, etc.)
- Use the `autocomplete` attribute to help browsers auto-fill fields
- In most cases use visually visible labels.

## DoR
- [x] Item has been estimated by the team
- [x] Item dependencies have been identified and documented

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

smfonseca:
[...]

PR #2095, opened 2025-03-31
fix: sd-input a11y reaudit

smfonseca:
## Description:
Closes [#2054](https://github.com/solid-design-system/solid/issues/2054)

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/input/input.ts
+++ b/packages/components/src/components/input/input.ts
@@ -646,7 +646,6 @@ export default class SdInput extends SolidElement implements SolidFormControl {
                     class=${cx('flex items-center sd-interactive', iconMarginLeft)}
                     type="button"
                     @click=${this.handlePasswordToggle}
-                    aria-pressed=${this.passwordVisible ? 'true' : 'false'}
                   >
                     ${this.passwordVisible
                       ? html`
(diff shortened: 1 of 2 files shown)


packages/components/src/components/input/input.ts
```diff
@@ -642,28 +643,20 @@ export default class SdInput extends SolidElement implements SolidFormControl {
                   <button
                     aria-label=${this.localize.term(this.passwordVisible ? 'hidePassword' : 'showPassword')}
                     part="password-toggle-button"
-                    class="flex items-center"
+                    class=${cx('flex items-center sd-interactive', iconMarginLeft)}
                     type="button"
                     @click=${this.handlePasswordToggle}
-                    tabindex="-1"
+                    aria-pressed=${this.passwordVisible ? 'true' : 'false'}
```
mariohamann, line 649: Does this work together with the current label? My expectation would be that then the label would have to stay the same, as now the following would happen when the password is visible:

- `label="hide password"` 
- `aria-pressed="true"`

Which doesn't really make sense, right?

(I know, that this propsal came from @matuzo, but I believe this combination doesn't make sense. So either we get rid of "aria-pressed" or we convert the label to being "non-ternary" and getting rid of "hidePassword".

mariohamann (approved):

mariohamann (approved):

packages/components/src/components/input/input.ts
smfonseca: Retested again and definetely does not make sense here, removed the aria attribute and kept the label which feels more informative IMO

paulovareiro29 (approved):

Issue #2064, opened 2025-03-21
feat[dev]: ✨ implement A11y Re-Improvements to sd-tooltip


yoezlem:
## Description
This tasks describes the Figma implementation for: #2048 

This task describes the Storybook implementation for the sd-tooltip component as part of the A11y re-improvements outlined in Epic #2048.

A11y improvements for sd-tooltip can be found here: [sd-tooltip Notes](https://www.notion.so/matuzo/sd-tooltip-1baf3f6e1bc081f186edd618c62683d7).

------------------------------------------

# New Comment
**Problem:** The button doesn’t indicate whether it’s open.
**Solution:** Put `aria-expanded` on the button and toggle it on click.

---

**Problem:** Focus is lost when you click a button. 
**Solution:** Keep focus on the button.

---

**Problem:** Buttons are not accessible with VoiceOVer in Safari 18.2 on macOS.
**Solution:** I'm not exactly sure what the reason is but it may have to do with you using `display: contents`.

## Comment
/

## DoR
- [x] Item has been estimated by the team
- [x] Item dependencies have been identified and documented

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #2138, opened 2025-04-14
fix: sd-tooltip a11y re-audit

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/2064

## Definition of Reviewable:


paulovareiro29 (changes requested):
@auroraVasconcelos  I have found a couple of issues:
- When closing the tooltip using keyboard, focus is not being kept on the button which means that focus is being lost.
- On Safari, with default trigger button, it is reading as "Tooltip tooltip" instead of "Tooltip Lorem Ipsum". I believe this is due to the extra `aria-label` on the `slot[name="content"]`.

auroraVasconcelos:
> @auroraVasconcelos I have found a couple of issues:
> 
> * When closing the tooltip using keyboard, focus is not being kept on the button which means that focus is being lost.

@paulovareiro29, on the Default we have a 'Click Focus' trigger. If the focus stays in the button after closing it, then you can't open the button with focus. Therefore I have only implemented the focus staying if, and only if, the trigger is 'Click'.

paulovareiro29:
> > @auroraVasconcelos I have found a couple of issues:
> > 
> > * When closing the tooltip using keyboard, focus is not being kept on the button which means that focus is being lost.
> 
> @paulovareiro29, on the Default we have a 'Click Focus' trigger. If the focus stays in the button after closing it, then you can't open the button with focus. Therefore I have only implemented the focus staying if, and only if, the trigger is 'Click'.

Aligned with Aurora that when using `Click Focus`, when using the keyboard after the tooltip is focused (and open), it should override the "opened by focus" and close the tooltip while keeping the trigger focused.

paulovareiro29 (changes requested):
In the `Custom Trigger` story it states:
> In this example we provide a solution to this problem by using a live region to announce the tooltip content when the trigger element is clicked.

I've tested it and it is does not announce anything. Could you have a look into it, since we are already taking care of improving the tooltip a11y.

paulovareiro29 (approved):

.changeset/public-parks-help.md
```diff
@@ -0,0 +1,10 @@
+---
+'@solid-design-system/components': minor
```
paulovareiro29, line 2:
```suggestion
'@solid-design-system/components': patch
```

paulovareiro29 (changes requested):
Change changeset to patch, since it is improvements, not new features

.changeset/public-parks-help.md
paulovareiro29: Also, create a `docs` changeset to inform the tooltip story was improved regarding a11y

paulovareiro29 (changes requested):

paulovareiro29 (approved):

smfonseca (approved):

MartaPintoTeixeira (approved):

PR #2140, opened 2025-04-15
fix: sd-dialog a11y reaudit

### LINKED ISSUE #2073: feat[dev]: ✨ implement A11y Re-Improvements to sd-dialog
## Description
This tasks describes the Figma implementation for: #2048 

This task describes the Storybook implementation for the sd-dialog component as part of the A11y re-improvements outlined in Epic #2048.

A11y improvements for sd-dialog can be found here: [sd-dialog Notes](https://www.notion.so/matuzo/sd-dialog-1baf3f6e1bc0819199f6f28805070d1d).

------------------------------------------

## Comment
/

## Notes
- The UX in timed dialogs would be better if the disabled button was focusable (see [[sd-button](https://www.notion.so/sd-button-1baf3f6e1bc08198a2d6ddf4df8b1e21?pvs=21)](https://www.notion.so/sd-button-1baf3f6e1bc08198a2d6ddf4df8b1e21?pvs=21)). You could use `aria-describedby` on the button and create a reference to the countdown. That way, the user would know why the button is disabled.

```html
<div id="countdown">
  Closable in 4 seconds...
</div>

<button aria-disabled="true" aria-describedby="countdown">
	Close
</button>
```

This example is currently on under the sd-dialog screenshots "Preventing closing".

**Solution:**
- Move the "Preventing closing" screenshot to a template and implement it using correct accessibility patterns. (Visually disabled, aria-describedby..)

- Please re-check the `aria-label` and `aria-labelledby`. Whenever the `headline` attribute is not set, check if aria-label is still rendered and if it being overriding `aria-labelledby`. It shouldn't be the case.
When the `headline` attribute is not set, aria-labelledby should be the one announced.

## Accessibility information
- Always provide a headline for the dialog.
- *Move the keyboard focus to the first interactive element within the modal (usually the close button) when the modal opens. Keep the keyboard focus within the modal while it is open, and return focus to the element that triggered the modal when it is closed.* **(To be removed on figma & overview pages)**

## DoR

## DoD

paulovareiro29:
## Description:

Closes #2073

## Definition of Reviewable:


paulovareiro29:
@coraliefeil @MartaPintoTeixeira 
I have updated the `sd-dialog` overview page by adding a known browser issue warning and removing some unecessary a11y information for our users. [Quick link to overview page](https://solid-design-system.github.io/solid/feat_overview-page-for-all-components/?path=/docs/components-sd-dialog-overview--docs#accessibility-information)

In this PR a new template story has been added as well. [Quick link to prevent closing template](https://solid-design-system.github.io/solid/fix_sd-dialog-a11y-reaudit/?path=/docs/templates-dialog--docs#prevent-closing)

auroraVasconcelos (dismissed):

MartaPintoTeixeira:
on the screenshot test can we name the buttons as "button" instead of "label" i find is misleading.

MartaPintoTeixeira:
Why do we need a "default" story in dialog Docs?
it is exactly the same as the 1st sample of the component and we never do this for other components

packages/docs/src/stories/templates/dialog.stories.ts
```diff
@@ -200,3 +201,98 @@ export const Scrollable = {
...
+        </script>
+
+        <sd-dialog id="timed" headline="Timed Dialog" no-close-button>
+          <div id="countdown">Closable in 5 seconds...</div>
+          <sd-button slot="footer" class="w-full" id="timed-close-button" visually-disabled>
+            Close <span id="close-countdown" class="sr-only">. Closable in 5 seconds...</span>
+          </sd-button>
+        </sd-dialog>
+
+        <script>
```
smfonseca, line 251: At the moment there is no announcement when the dialog can be closed. Could be a nice addition to also implement this.

MartaPintoTeixeira:
Dialog new tmplt in figma.

packages/docs/src/stories/templates/dialog.stories.ts
paulovareiro29: @smfonseca If the close button is focused, this is already happening.
Do you think it is necessary to do a live announcement that the dialog is closable now, even if the close button is not focused? (there is a clear intent to close).

smfonseca: If you open the dialog you can tab into the close button and it's announced that it will close in 5 seconds (even it's only 2s). When the time elapses and the button is available, there is no announcement. Only if I tab away from the button and back to it will I now that I can now close the dialog. In this scenario I would like to have the announcement. What do you think?

paulovareiro29: I just tested on different browsers and here is the result:
- On Chrome, if I keep the button focused, when the time elapses and the button becomes available, it announces that it is closable.
- On Safari, it doesn't announce anything while we keep the button focused.

IMO, this is a browser issue that Safari needs to improve, since Chrome already does it. What do you think?

smfonseca: Ok, I only tested in Safari! 
Let's go with what we currently have.
Thanks for checking on both browsers.

smfonseca (approved):

Issue #2113, opened 2025-04-08
fix!: 🤔 `sd-scrollable` `start` and `end` events are not working as expected (BREAKING CHANGE)


paulovareiro29:
## Current behavior

**_BREAKING CHANGE_**

In the `sd-scrollable`, two events are not working as expected:
`start`: Should be "Emitted when the start of the scrollable is reached."
`end`: Should be "Emitted when the end of the scrollable is reached."

In reality, these events are being triggered when:
`start`: There is available scrolling space to the start direction
`end`: There is available scrolling space to the end direction

This means that these events are called a lot of time when the container is being scrolled and even both events are called when the scroll position is in the middle.

## Expected behavior
`start`: Should be emitted when the start of the scrollable is reached.
`end`: Should be emitted when the end of the scrollable is reached.

## Technical Information

**_This will introduce a breaking change_**

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #2209, opened 2025-05-14
fix!: sd-scrollable start and end events

paulovareiro29:
## Description:

Addresses #2113 

## Definition of Reviewable:

### DIFF
--- a/.changeset/lucky-camels-cut.md
+++ b/.changeset/lucky-camels-cut.md
@@ -2,8 +2,6 @@
 '@solid-design-system/components': major
 ---
 
-`sd-scrollable` `start` and `end` events were not working as expected.
-- `start` event was emitted when there was available scrolling space in the start direction.
-- `end` event was emitted when there was available scrolling space in the end direction.
-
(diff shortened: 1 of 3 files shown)


.changeset/lucky-camels-cut.md
```diff
@@ -0,0 +1,9 @@
+---
+'@solid-design-system/components': major
+---
+
+`sd-scrollable` `start` and `end` events were not working as expected.
+- `start` event was emitted when there was available scrolling space in the start direction.
+- `end` event was emitted when there was available scrolling space in the end direction.
+
+Now events are working as expected, only being triggered when the `start`/`end` is reached.
```
smfonseca, line 9: I would remove the first phrase and adopt our usual writing style of "fixed....". If we are doing a fix it's because something was not working as expected.
And for the the events I think it would be nice to document it as a before -> after. 

Example:
```
- (before): `start` event emitted when... ---> (after): `start` event emmitted when...
```

What do you think?

paulovareiro29: Sounds good!

Issue #1960, opened 2025-03-11
fix: 🤔 sd-input Read Only text color


MartaPintoTeixeira:
## Current behavior
Read only sd-input sample is not the same in figma and in storybook.
In storybook: the font color of the copy inside the input "Read only example" should be "filled" (text/black) as it is in figma.

Figma:
<img width="830" alt="Image" src="https://github.com/user-attachments/assets/7004f447-b88f-4d2d-8312-9937093f8223" />

Storybook:
<img width="1373" alt="Image" src="https://github.com/user-attachments/assets/0fb5f46a-ba48-4b31-81be-29d048e979db" />

## Expected behavior
"Read only example" text should be "black" as in figma.

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

balco0110:
Hey @MartaPintoTeixeira 😄 i think this is alredy done, can you pls check and let me know

PR #2213, opened 2025-05-16
Changed the color of the Read only input text

balco0110:
-
-
-

## Description:

https://github.com/solid-design-system/solid/issues/1960

## Definition of Reviewable:

### DIFF
--- a/.changeset/tired-apples-rule.md
+++ b/.changeset/tired-apples-rule.md
@@ -1,0 +1,5 @@
+---
+'@solid-design-system/docs': patch
+---
+
+Changed the color of the read only input text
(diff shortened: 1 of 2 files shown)


.changeset/twelve-rockets-run.md
paulovareiro29, line 1: Lets just have one changeset, otherwise two patch versions will be created. Let's delete this one, since the other has a nicer text

paulovareiro29 (changes requested):

.changeset/tired-apples-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/docs': patch
+---
+
+Changed the color of the read only input text
```
paulovareiro29, line 5: Since we didn't actually change any colors, I suggest we change the changeset to better indicate what was done.

```suggestion
Updated the input read-only story to use `value` instead of `placeholder`.
```

paulovareiro29 (changes requested):

balco0110 (commented):
updated the coment in the changeset

PR #2256, opened 2025-05-29
feat: add motion design to sd-tab

smfonseca:
## Description:
Closes https://github.com/solid-design-system/solid/issues/2194

## Definition of Reviewable:


packages/components/src/components/tab-group/tab-group.ts
```diff
@@ -303,6 +306,30 @@ export default class SdTabGroup extends SolidElement {
     this.hasScrollControls = this.nav.scrollWidth > this.nav.clientWidth;
   }
 
+  private async repositionIndicator() {
+    await this.updateComplete;
+
+    const currentTab = this.activeTab;
+    const indicator = this.renderRoot.querySelector<HTMLElement>('#indicator');
+    const tabsContainer = this.renderRoot.querySelector<HTMLElement>('[part="tabs"]');
```
smfonseca: I think it's best to move this query to `firstUpdated` or instead use a getter function.

[https://lit.dev/docs/components/shadow-dom/#accessing-nodes-in-the-shadow-dom](https://lit.dev/docs/components/shadow-dom/#accessing-nodes-in-the-shadow-dom)

packages/components/src/components/tab-group/tab-group.ts
```diff
@@ -348,7 +375,13 @@ export default class SdTabGroup extends SolidElement {
 
           <div part="scroll-container" class="flex overflow-x-auto focus-visible:focus-outline !outline-offset-0">
             <div part="tabs" class=${cx('flex flex-auto relative flex-row')} role="tablist">
-              <div part="separation" class="w-full border-[0.25px] border-neutral-400 absolute bottom-0"></div>
+              <div part="separation" class="border-neutral-400 absolute w-full h-0.25 bottom-0 border-b"></div>
+              <div
+                part="active-tab-indicator"
+                id="indicator"
+                class="absolute h-1 bg-accent bottom-0 transition-all duration-medium ease-in-out z-10"
```
smfonseca: `transition-[transform,width]` should be enough in this case, please double check

packages/components/src/components/tab-group/tab-group.ts
```diff
@@ -303,6 +306,30 @@ export default class SdTabGroup extends SolidElement {
...
+    const currentTab = this.activeTab;
+    const indicator = this.renderRoot.querySelector<HTMLElement>('#indicator');
+    const tabsContainer = this.renderRoot.querySelector<HTMLElement>('[part="tabs"]');
+
+    if (!currentTab || !indicator || !tabsContainer) return;
+
+    if (currentTab.variant === 'container') {
+      indicator.style.display = 'none';
+      return;
+    }
```
smfonseca: You can remove this and add `currentTab.variant === 'container'` on L316

packages/components/src/components/tab/tab.ts
```diff
@@ -109,9 +121,8 @@ export default class SdTab extends SolidElement {
           <div
```
smfonseca: Only render this div if variant is container

packages/components/src/components/tab-group/tab-group.ts
```diff
@@ -348,7 +375,13 @@ export default class SdTabGroup extends SolidElement {
 
           <div part="scroll-container" class="flex overflow-x-auto focus-visible:focus-outline !outline-offset-0">
             <div part="tabs" class=${cx('flex flex-auto relative flex-row')} role="tablist">
-              <div part="separation" class="w-full border-[0.25px] border-neutral-400 absolute bottom-0"></div>
+              <div part="separation" class="border-neutral-400 absolute w-full h-0.25 bottom-0 border-b"></div>
+              <div
```
smfonseca: Let's only render this if the tabs are variant default.

packages/components/src/components/tab-group/tab-group.ts
auroraVasconcelos: Updated

packages/components/src/components/tab-group/tab-group.ts
auroraVasconcelos: Updated

packages/components/src/components/tab/tab.ts
auroraVasconcelos: Updated

packages/components/src/components/tab-group/tab-group.ts
auroraVasconcelos: Updated

paulovareiro29 (changes requested):
`Tab Group Template` > `Tab Group Center Aligned` is broken

packages/components/src/components/tab-group/tab-group.ts
```diff
@@ -47,6 +47,8 @@ export default class SdTabGroup extends SolidElement {
   @query('[part=base]') tabGroup: HTMLElement;
   @query('[part=body]') body: HTMLSlotElement;
   @query('[part=scroll-container]') nav: HTMLElement;
+  @query('#indicator') private indicator!: HTMLElement;
```
paulovareiro29:
```suggestion
  @query('[part=active-tab-indicator]') indicator: HTMLElement | null;
```

Any reason to use an extra `#id` instead of the part?

packages/components/src/components/tab/tab.ts
```diff
@@ -52,8 +57,15 @@ export default class SdTab extends SolidElement {
...
-    this.setAttribute('aria-selected', this.active ? 'true' : 'false');
+  async handleActiveChange() {
+    if (this.variant === 'container' && this.active) {
+      await stopAnimations(this.tab);
+
+      const animation = getAnimation(this, 'containerTab.active', { dir: this.localize.dir() });
+      await animateTo(this.tab, animation.keyframes, animation.options);
+    }
+
+    return this.setAttribute('aria-selected', this.active ? 'true' : 'false');
```
paulovareiro29: This returns `undefined`. No need to return anything here.

```suggestion
   this.setAttribute('aria-selected', this.active ? 'true' : 'false');
```

packages/components/src/components/tab/tab.ts
```diff
@@ -145,6 +160,11 @@ export default class SdTab extends SolidElement {
   ];
 }
 
+setDefaultAnimation('containerTab.active', {
+  keyframes: [{ opacity: 0 }, { opacity: 1 }],
+  options: { duration: 'var(--sd-duration-medium, 300)', easing: 'ease-in-out' }
+});
+
```
paulovareiro29: The animation doesn't feel correct.
In the docs it states that the opacity is only on the border, and not on the whole tab.
With current implementation the tab is flickering, when tab is `variant="container"`

MartaPintoTeixeira (commented):
tmplt Tab Group with Icon and Badge:
- Is the size of the dashboard icon in the same as the ones next to it?
- This template is broken when is comes to spacing. Can you check figma?

For the motion please check the tab with contour - it should have a quick fade it.

auroraVasconcelos:
> tmplt Tab Group with Icon and Badge:
> 
> * Is the size of the dashboard icon in the same as the ones next to it?
> * This template is broken when is comes to spacing. Can you check figma?
> * Placement of the badge is wrong
> 
> For the motion please check the tab with contour - it should have a quick fade it.

To be addressed with https://github.com/solid-design-system/solid/issues/2142

packages/components/src/components/tab/tab.ts
auroraVasconcelos: Updated

paulovareiro29 (changes requested):
The motion of the sd-tab container variant kind of feels too quick, and I guess its because the green bar indicator appears without motion.

What do you think about adding the same motion to the indicator as we have in the `navigation-item`?

MartaPintoTeixeira (changes requested):
sd-tab docs
- visually disables sample: change copy sample to "Visually disabled"
- icon sample if different in figma
- active sample if different in figma

tab motion: isn't it too fast? can you double check speed? thanks

paulovareiro29 (changes requested):
Please add the changeset for the `docs` package.
Other than that, it looks good!

auroraVasconcelos:
> Please add the changeset for the `docs` package. Other than that, it looks good!

Updated

paulovareiro29 (dismissed):

paulovareiro29 (approved):

MartaPintoTeixeira (approved):

PR #2270, opened 2025-06-03
feat: add carousel fade effect

smfonseca:
## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/carousel/fade-controller.ts
+++ b/packages/components/src/components/carousel/fade-controller.ts
@@ -1,0 +1,202 @@
+import type SdCarousel from './carousel';
+
+export class FadeController {
+  private carousel: SdCarousel;
+  private heightCalculated = false;
+  private resizeTimeout?: ReturnType<typeof setTimeout>;
+  private isEnabled = false;
+  private isCalculating = false;
+
+  private readonly fadeClasses = {
+    base: [
+      'absolute',
+      'top-0',
+      'left-0',
+      'w-full',
+      'h-full',
+      'visible',
+      'transition-opacity',
+      'duration-x-slow',
+      'ease-in-out'
+    ],
+    active: ['opacity-100', 'z-[1]'],
+    inactive: ['opacity-0', 'z-0']
+  } as const;
+
+  constructor(carousel: SdCarousel) {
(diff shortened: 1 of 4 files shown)


MartaPintoTeixeira (approved):
carrousel fade effect is good.
but what about the dotted motion for carrousel? is that in another ticket?
if yes consider it approved

packages/components/src/components/carousel/fade-controller.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    if (isActive) {
+      classList.add(...this.fadeClasses.active);
+    } else {
+      classList.add(...this.fadeClasses.inactive);
+    }
+
+    slide.toggleAttribute('inert', !isActive);
+    slide.setAttribute('aria-hidden', isActive ? 'false' : 'true');
+  }
```
paulovareiro29: Probably here it is better to do a separation of concerns. We could create the `show` and `hide` methods.
This way we prevent from having unnecessary "isActive" parameter on methods, and necessary if conditions inside the method, which increases complexity.

Then on the `slides.forEach`, that's where we would add the condition if it is active.

```
if (index === activeIndex) {
 this.show(slide);
} else {
 this.hide(slide);
}
```

packages/components/src/components/carousel/fade-controller.ts
paulovareiro29, line 1: Should we add some unit tests here?

packages/components/src/components/carousel/fade-controller.ts
smfonseca: Updated

packages/components/src/components/carousel/fade-controller.ts
smfonseca: Added

smfonseca:
> carrousel fade effect is good. but what about the dotted motion for carrousel? is that in another ticket? if yes consider it approved

Has mentioned in the daily, dots animation was done before. PR can be found [here](https://github.com/solid-design-system/solid/pull/2219).

paulovareiro29 (approved):

MartaPintoTeixeira (approved):

Issue #1312, opened 2024-08-23
fix: 🤔 Unable to change font-size for headline in sd-teaser-media


auroraVasconcelos:
## Current behavior
When using the sd-teaser-media is is not possible to adjust the font size the slot `headline`.

## Expected behavior
`Headline` size can be changed as desired.

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

Issue #1558, opened 2024-10-15
fix: 🤔 unable to change headline margins or size in teaser


auroraVasconcelos:
## Current behavior
It is not possible to change the font size or the margins of the headline in the teaser.

## Expected behavior
Considering the different layouts a teaser could have (check [figma](https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=4187-6785&t=oFhf4IDDB5DDa6qK-4)) the headline styles needs to be addaptable.

## Steps to reproduce
Create a sd-teaser with a slot headline and try to add a margin or to change the font-size.

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

mariohamann:
https://codepen.io/mariohamann/pen/RwXVabR?editors=1110

It's possible by wrapping. Maybe it's something with the specifity (weird, that ::slotted is so nasty). If it prevents you from doing the teaser templates, go the route of wrapping elements first.

PR #2405, opened 2025-07-28
fix: removed the fixed header size and margin

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/1312 and https://github.com/solid-design-system/solid/issues/1558
## Definition of Reviewable:


smfonseca (changes requested):
This is a breaking change, do not merge.

balco0110:
> > Do you think this is a breaking change @smfonseca ? It will break the UI for some teasers
> 
> Yes, I believe so.
> 
> @mariohamann left a suggestion to address this issue [here](https://github.com/solid-design-system/solid/issues/1558#issuecomment-2415727045).
> 
> In his example he's targeting directly the `h4`, but this will cascade to all other teasers. To make it more precise, I suggest adding a class to the `headline` slot and everything should work as expected.
> 
> Can you try it @balco0110?

i've pushed the code, added a class and removed the ::slotted([slot='headline']) part, in this way we are able to change the size of the text

paulovareiro29:
Converting to draft to prevent from merging

MartaPintoTeixeira (approved):
Approving as it is not a visual change

mariohamann:
I think this whole thing needs some re-consideration. Ideally it should work without a breaking change. Is this possible in any way?

packages/components/src/components/teaser-media/teaser-media.ts
```diff
@@ -83,7 +83,7 @@ export default class SdTeaserMedia extends SolidElement {
             <div class="flex-col text-left p-4" part="content">
               <div class="flex flex-col">
                 <div part="headline" class="text-lg font-bold m-0 order-2">
-                  <slot name="headline"
+                  <slot name="headline" class="@apply font-bold !m-0 !text-lg;"
```
mariohamann: `@apply` is not a class, `!text-lg;` as well. Mind your copy paste. ;)

PR #2457, opened 2025-08-11
fix: sd-step horizontal inline

auroraVasconcelos:
## Description:
- Removes extra space on the last sd-step of a step-group

## Definition of Reviewable:


van-nguyen-ht:
I have checked the storybook on the text vertically aligned to the number in the inline variant (last step) and I approve the visual adjustment.

packages/components/src/components/step/step.ts
```diff
@@ -219,35 +219,33 @@ export default class SdStep extends SolidElement {
...
                       )}
                     >
-                      ${hasLabel
+                      ${hasLabel &&
+                      html`
+                        <div
+                          part="label"
+                          id="label"
+                          class=${cx(
+                            '!font-bold sd-paragraph whitespace-nowrap',
```
smfonseca:
```suggestion
                            'font-bold sd-paragraph whitespace-nowrap',
```

packages/components/src/components/step/step.ts
```diff
@@ -219,35 +219,33 @@ export default class SdStep extends SolidElement {
...
+                            this.disabled && '!text-neutral-500',
+                            this.waiting && '!text-neutral-700',
```
smfonseca: Please double-check if the !important is really necessary, ideally we should avoid using it. Same for L240 and L269

packages/components/src/components/step/step.ts
auroraVasconcelos: Updated

packages/components/src/components/step/step.ts
auroraVasconcelos: Updated

packages/components/src/components/step/step.ts
```diff
@@ -321,8 +321,8 @@ export default class SdStep extends SolidElement {
                     id="description"
                     class=${cx(
                       'sd-paragraph sd-paragraph--size-sm',
-                      this.disabled && '!text-neutral-700',
-                      this.waiting && '!text-neutral-700'
+                      this.disabled && 'text-neutral-700',
+                      this.waiting && 'text-neutral-700'
```
smfonseca: This can be combined into `(this.disabled || this.waiting) && 'text-neutral-700'`

auroraVasconcelos: Updated

Issue #2153, opened 2025-04-17
fix: 🤔 `sd-dialog` template unwanted scrolling


paulovareiro29:
## Current behavior
When closing a dialog in the `sd-dialog` templates, sometimes the page scrolls to another template in which the dialog is open.

## Expected behavior
The page should stay where it is, meaning that it should not scroll to another story.

## Steps to reproduce

1. Open the `sd-dialog` templates.
2. Click "Open Dialog" on the "Prevent Closing" story.
3. Close the dialog.

Sometimes it scrolls up to another story. It doesn't always happen.

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

PR #2466, opened 2025-08-19
fix: sd-dialog unwanted scrolling

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/2153

## Definition of Reviewable:


packages/components/src/components/dialog/dialog.ts
```diff
@@ -228,8 +228,15 @@ export default class SdDialog extends SolidElement {
...
-      if (typeof trigger?.focus === 'function') {
-        setTimeout(() => trigger.focus());
+      if (typeof trigger?.focus === 'function' && trigger.isConnected) {
+        setTimeout(() => {
+          try {
+            trigger.focus({ preventScroll: true });
+          } catch {
+            trigger.focus();
+          }
+          this.originalTrigger = null;
```
smfonseca, line 238: Why do we need to reset the trigger? Is there a specific reason for this?

.changeset/afraid-clubs-give.md
```diff
@@ -0,0 +1,7 @@
+---
+'@solid-design-system/components': patch
+---
+
+Improve sd-dialog:
+
+- Prevent unexpected scrolling behavior.
```
smfonseca:
```suggestion
- Prevent unexpected scrolling behaviour.
```

smfonseca (commented):
Can

packages/components/src/components/dialog/dialog.ts
auroraVasconcelos: The goal is to prevent it to focus on an old element when closing the dialog

smfonseca: I think `this.originalTrigger` doesn't change, it's always the same element, that's why I'm not sure if this is needed. If you tested that this doesn't cause any problem, then I'm fine with you moving this PR forward.

smfonseca: Discussed in teams.

Issue #2512, opened 2025-09-18
chore: 🔧 set `minimumReleaseAge` in `pnpm` to 5 days


mariohamann:
## User Story

As a maintainer I want to improve security for us and our users.

### Suggested Solution

- [ ] Update version via core pack to 10.16
- [ ] Set `"pnpm": "^10.0.0"` in `package.json/engines` to at least `10.16`
- [ ] Set https://pnpm.io/settings#minimumreleaseage to 5 days
- [ ] Update all actions, to ensure that they're using the correct version

## Environment (GitHub Actions or Azure DevOps)

GitHub Actions, package.json

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

karlbaumhauer:
@solid-design-system/development pls make a small dev-only-refinement on this one. As we wont have a refinement, this should not wait until next week.
clarified with @yoezlem

auroraVasconcelos:
The Solid, Communications, and Showcase repositories in GitHub, as well as the pipelines in Azure, were updated. The demo repository was skipped due to inactivity.

PR #2522, opened 2025-10-01
chore: update dependencies & set minimumReleaseAge in pnpm to 5 days

auroraVasconcelos:
## Description:
Closes https://github.com/solid-design-system/solid/issues/460, closes https://github.com/solid-design-system/solid/issues/2512

## Definition of Reviewable:


packages/components/src/components/input/input.ts
```diff
@@ -632,7 +632,7 @@ export default class SdInput extends SolidElement implements SolidFormControl {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
                 ifDefined(this.autocomplete as any)
               }
-              autocorrect=${ifDefined(this.type === 'password' ? 'off' : this.autocorrect)}
+              autocorrect=${ifDefined(this.type === 'password' ? 'off' : this.autocorrect ? 'on' : 'off')}
```
mariohamann: We need to keep the original types here, otherwise we have a breaking change.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocorrect

mariohamann (approved):

PR #2567, opened 2025-10-28
fix: 🐛 fix broken builds

mariohamann:
## Description:

Address issues causing build failures by fixing broken imports and ensuring empty files are populated with placeholder content. This is NOT a fix of the root cause, which is still unclear, but it works.

This simply fills empty JS files with a comment, so that its import works:

Before:

```
```

After:

```js
//
```

## Definition of Reviewable:

### DIFF
--- a/.changeset/six-walls-search.md
+++ b/.changeset/six-walls-search.md
@@ -2,4 +2,4 @@
 '@solid-design-system/components': patch
 ---
 
-Fix broken build
+Fixed issue where the build process would generate empty chunk files, therefore, breaking the build.


smfonseca (dismissed):

.changeset/six-walls-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/components': patch
+---
+
+Fix broken build
```
paulovareiro29:
```suggestion
Fixed issue where the build process would generate empty chunk files, therefore, breaking the build.
```

packages/components/scripts/build.js
```diff
@@ -164,6 +164,18 @@ async function runBuild() {
     buildResults = await buildTheSource();
   });
 
+  await nextTask('Fix broken imports', async () => {
+    const distFiles = await globby([`${outdir}/chunks/chunk.*.js`, `${cdndir}/chunks/chunk.*.js`]);
```
paulovareiro29, line 168: Don't we need to add the versioned folders as well?

PR #2623, opened 2025-11-17
fix: refactor counter logic for ordered lists

Vahid1919:
## Description:

Replace the use of counters() with counter(). counters() is unnecessary here since we explicitly define three variables anyway. This occasionally caused issues in Chrome when multiple lists were present in CMS test pages. Switching to counter() avoids these edge-case bugs and makes the code more readable.

## Definition of Reviewable:

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,5 +1,15 @@
 # @solid-design-system/components
 
+## 5.17.0
+
+### Minor Changes
+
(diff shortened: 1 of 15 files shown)


.changeset/sour-books-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/styles': patch
+---
+
+Replaced counters() with counter() inside ordered lists.
```
mariohamann: Please provide an info why this was done.

Issue #2372, opened 2025-07-18
fix: 🤔 sd-menu – mouse click hinders keyboard behaviour


mariohamann:
## Current behavior

When I click on a menu button with the mouse, I can't interact with the keyboard afterwards.

https://github.com/user-attachments/assets/23ed69b7-9476-4ca1-8601-5016cc8cc3e2

## Expected behavior

It should have the same behaviour as when clicking with the enter button.

## Steps to reproduce

https://codepen.io/mariohamann/pen/OPNzWBK?editors=1110:
- Open Menu with Tab and Return: Keyboard works
- Open Menu with Mouse: Keyboard doesn't work

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

yoezlem:
@martascorreia can we close the ticket or is something missing here?

PR #2658, opened 2025-12-12
fix: mouse/keyboard interactions of menu/menu-item (DELETE)

martascorreia:
## Description:
Closes https://github.com/solid-design-system/solid/issues/2372. The problem is related to activation of focus-visible. 

When we click the dropdown menu with the mouse, the focus is put on the menu-item correctly with "focus", however, the focus-visible that puts the border around the item is not activated. This property only activates after a keyboard event. So for example in Safari, if you do "arrowdown" and then click "arrowright" (or any other key we are not catching), the outline suddenly appears because an event key was propagated. 

My solution was to force this css to appear, and also remove it when not needed. Please let me know if you know a better way to handle this.

## Definition of Reviewable:
- [X] relevant tickets are linked

### DIFF
--- a/.changeset/tough-flies-think.md
+++ b/.changeset/tough-flies-think.md
@@ -2,4 +2,4 @@
 '@solid-design-system/components': patch
 ---
 
-fixing mouse click then keyboard use after
+Fixed `sd-menu` keyboard interaction issue after enabling it with a mouse click.


.changeset/tough-flies-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/components': patch
+---
+
+fixing mouse click then keyboard use after
```
smfonseca:
```suggestion
Fixed `sd-menu` keyboard interaction issue after enabling it with a mouse click.
```

packages/components/src/components/dropdown/dropdown.ts
```diff
@@ -269,6 +269,13 @@ export default class SdDropdown extends SolidElement {
 
   handleTriggerClick() {
     if (this.open) {
+      const menu = this.getMenu();
+      if (menu) {
+        const current = menu.getCurrentItem();
+        current?.classList.remove('force-visible-focus');
+        console.log(current);
```
smfonseca: Please remove log

packages/components/src/components/menu-item/menu-item.ts
```diff
@@ -229,6 +229,10 @@ export default class SdMenuItem extends SolidElement {
         @apply focus-outline outline-offset-[-2px];
       }
 
+      :host(.force-visible-focus) [part='base'] {
```
smfonseca: I would prefer not to use the word `force`. Either remove it or call it something like `menu-item-focus`.

Issue #2295, opened 2025-06-17
fix: 🤔 tooltip should close when clicking outside element


smfonseca:
## Current behavior
Tooltip is not closable when clicking outside the element.
This is happening in Safari (v18.5), works ok on Chrome.

## Expected behavior
Tooltip closes when clicking somewhere else on the page.

## Steps to reproduce
Open the [tooltip in the docs](https://solid-design-system.fe.union-investment.de/docs/?path=/docs/components-sd-tooltip--docs) using Safari and try to close it when clicking somewhere else.

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] Documentation has been created/updated (if applicable)
- [ ] Migration Guide has been created/updated (if applicable)
- [ ] Relevant E2E tests (Features, A11y, Bug fixes) are created/updated
- [ ] Relevant stories (Features, A11y) are created/updated
- [ ] Implementation works successfully on `feature` branch

martascorreia:
Review from devs: Needs tests.

PR #2661, opened 2025-12-16
fix: tooltip not closing outside of element

martascorreia:
## Description:
  Closes https://github.com/solid-design-system/solid/issues/2295. 

A new attribute "close-trigger" was added to the tooltip, mirroring the current attribute "trigger" that now works more as a "open-trigger" (I didn't change the name, should I?). This "close-trigger" defines which triggers should make the tooltip close. For the specific case of sd-range where the tooltip should always stay open, the sd-range changes the "close-trigger" to manual.

## Definition of Reviewable:

### DIFF
--- a/.changeset/pink-zebras-dream.md
+++ b/.changeset/pink-zebras-dream.md
@@ -1,9 +1,5 @@
 ---
 '@solid-design-system/components': patch
-'@solid-design-system/theming': patch
-'@solid-design-system/styles': patch
-'@solid-design-system/tokens': patch
-'@solid-design-system/docs': patch
 ---
 
 fixing tooltip closing outside of element + implementing a closing trigger atribute


.changeset/pink-zebras-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/components': patch
+---
+
+fixing tooltip closing outside of element + implementing a closing trigger atribute
```
smfonseca, line 5: Please describe a bit more the issue and the new attribute. When mentioning a component always add the `sd-` prefix.

packages/components/src/components/tooltip/tooltip.test.ts
```diff
@@ -205,7 +205,7 @@ describe('<sd-tooltip>', () => {
 
   it('should toggle the tooltip when focus the trigger', async () => {
     const el = await fixture<SdTooltip>(html`
-      <sd-tooltip content="This is a tooltip" trigger="focus">
+      <sd-tooltip content="This is a tooltip" trigger="focus" close-trigger="focus">
```
mariohamann, line 208: As this is the default – is this really needed?

packages/components/src/components/tooltip/tooltip.test.ts
```diff
@@ -223,7 +223,7 @@ describe('<sd-tooltip>', () => {
     // tooltip is visible after focusing the button
     expect(body.hidden).to.be.false;
 
-    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab', bubbles: true }));
     button.blur();
     await waitUntil(() => body.hidden);
```
mariohamann, line 228: Add tests for the triggers.

PR #2666, opened 2025-12-22
fix: sd select dropdown resizing (DELETE)

martascorreia:
## Description:
Closes: https://github.com/orgs/solid-design-system/projects/1/views/68?pane=issue&itemId=145154220&issue=solid-design-system%7Csolid%7C2660

Fixed the issue with the sd-select not resizing properly. The option drop down was also changed to wrap the words to the next line. The tags when inside the sd-select are also now handled and three points are added when the word is too big. Example:

image

## Definition of Reviewable:
- [X] relevant tickets are linked


mariohamann (approved):

MartaPintoTeixeira (changes requested):
sd-select screenshots:
- Size x Multiple: the "X" inside the option is not vertically centered.
- Valid x Invalid: if i uncheck the options on "true" the "Please select" is cutted. also if then i check again one option the size of the sd-select changes on the width with the error on and off and it should always be the same.

Screenshot 2026-01-20 at 11 51 57

Issue #2398, opened 2025-07-25
fix: 🤔 disallow `removable` on `sd-tag` when it has `toggleable` or `selected`


paulovareiro29:
## Current behavior
On the `sd-tag`, there is the possibility for it to have the attribute `removable` and also for it to be `toggleable`. This results in the sd-tag to be a `<button>` with a `<button>` inside, which is not accessible and doesn't make much sense.

## Expected behavior
Whenever a `sd-tag` has the attribute `toggleable` or `selected`, the removable attribute has no effect on it.

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

PR #2682, opened 2026-01-07
fix: disabled removabled when selected or toggleable

martascorreia:
## Description:
Closes: https://github.com/solid-design-system/solid/issues/2398.

Added a variable _isRemovable_ to the sd-tag component. This variable represents the _removable_ attribute being enabled, while checking if the  _toggleable_ or _selected_ attributes are disabled. Then, replaced every place that checked if the removable attribute was enabled for this new variable.

Also added a note to the docs about this. Let me know if it is enough or if we should create more emphasis. 

## Definition of Reviewable:


mariohamann (changes requested):
Add tests. (Just that there is only one button when removable and selected is active or similar.)

PR #2709, opened 2026-01-13
fix: 🐛 improve nxt docs and scripts

mariohamann:
## Description:
- Updated docs
- made .env available to set Figma file and key in a file
- clean up existing cdn and dist folder on building

## Definition of Reviewable:


packages/docs/src/stories/packages/Quickstart.mdx
smfonseca: @mariohamann we should also update this

Issue #2084, opened 2025-03-25
feat: ✨ Implementation of the warning sign in sd-checkbox + sd-checkbox-group + sd-radio-group


MartaPintoTeixeira:
## User Story
As a user of the Solid Design System, I would like to see a warning sign next to the error message, as well as ensure the code accurately reflects the implementation already done in Figma, so that accessibility requirements are met, and users can easily identify and understand errors.

See figma:
sd-checkbox
sd-checkbox-group
sd-radio-group

## DoR
- [ ] Item has business value
- [ ] Item all subtasks have been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] All subtasks have been closed

MartaPintoTeixeira:
In docs (acessibility) make sure to rephace existing error text bullet with the one bellow:
"An error-text with a warning icon should be placed underneath an invalid checkbox or, if used in a group, underneath the checkbox group. Error messages should always provide hints for solutions."

PR #2639, opened 2025-11-24
feat: add warning sign to text error (DELETE)

martascorreia:
## Description:
 Closes: https://github.com/solid-design-system/solid/issues/2084
- Adding the warning sign to the text error message for sd-checkbox, sd-switch and sd-radio-group;
- Changed the template for "Checkbox Group" in "Required Checkbox Group" so that it also hides the warning sign under each individual checkbox;
- Fixed the issue where the text error would appear on the same line as the main text, which was visible in Templates -> Forms -> Contact Form -> "I accept the Privacy Policy. *" after the error is activated.

## Definition of Reviewable:

- [ ] Documentation is created/updated

- [ ] Stories (features, a11y) are created/updated

PR #2745, opened 2026-01-20
fix: adding warning sign to text error

martascorreia:
## Description:
 Closes: https://github.com/solid-design-system/solid/issues/2084
Previously this PR: https://github.com/solid-design-system/solid/pull/2639

- Adding the warning sign to the text error message for sd-checkbox, sd-switch and sd-radio-group;
- Changed the template for "Checkbox Group" in "Required Checkbox Group" so that it also hides the warning sign under each individual checkbox;
- Fixed the issue where the text error would appear on the same line as the main text, which was visible in Templates -> Forms -> Contact Form -> "I accept the Privacy Policy. *" after the error is activated.

## Definition of Reviewable:

### DIFF
--- a/packages/components/src/components/checkbox/checkbox.ts
+++ b/packages/components/src/components/checkbox/checkbox.ts
@@ -330,7 +330,11 @@ export default class SdCheckbox extends SolidElement implements SolidFormControl
             <slot></slot>
           </span>
         </label>
-        <div id="invalid-icon-message" part="invalid-icon-message" class="flex items-center gap-2 mt-2">
+        <div
+          id="invalid-icon-message"
+          part="invalid-icon-message"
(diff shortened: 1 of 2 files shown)


MartaPintoTeixeira (changes requested):
sd-radio-group screenshots
- invalid: when user selects one of the Radios it should remove the error state. (see invalid checkbox as example)

checkbox-group template
- the template "Required Checkbox Group" is missing the error icons when i click in Submit

forms template
- the template "Contact Form" is broken on the checkbox. the error should come bellow (not on the side). and the error should have a warning sign.
- the template "Form With Visually Disabled Elements" has the same issue as above. if i check the "I accept..." checkbox and then uncheck it the error should come bellow (not on the side). and the error should have a warning sign.

Issue #2508, opened 2025-09-15
feat[dev]: ✨ add expand button to sd-teaser-media expandable slot


MartaPintoTeixeira:
## Description
Dev ticket after:
feat[design]: ✨ add expand button to sd-teaser-media expandable slot [#2320](https://github.com/solid-design-system/solid/issues/2320)

- Teaser Media now has an expandable atribute in figma and this must be synced with code
- New template: Teaser Media Expandable
- In overview docs add a bullet to "content":
"When using the expandable slot make sure it does not contain any action since it is not shown on mobile devices due to hover interaction. The slot should only conceal supplementary, non-essential content."

### Open Questions towards design
- [ ] Question1
- [ ] Question2

## DoR
- [ ] Item has been estimated by the team
- [ ] Item dependencies have been identified and documented

PR #2645, opened 2025-11-27
feat: add expand button to the sd-teaser media(DELETE)

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/2508
Previous was: https://github.com/solid-design-system/solid/pull/2757
## Definition of Reviewable:

- [x] Documentation is created/updated

- [x] Stories (features, a11y) are created/updated
- [x] relevant tickets are linked

PR #2757, opened 2026-01-21
feat: Implemented expand button on the `sd-teaser-media` and updated …

balco0110:
…the template according to figma

## Description:

Closes: https://github.com/solid-design-system/solid/issues/2508

## Definition of Reviewable:

### DIFF
--- a/packages/docs/src/stories/components/teaser-media.stories.ts
+++ b/packages/docs/src/stories/components/teaser-media.stories.ts
@@ -178,9 +178,9 @@ export const MetaSlot = {
 };
 
 /**
- * Use the `expandable slot` to add content that only shows up on hover. This slot cannot contain any action since it is not shown on mobile devices due to hover interaction.
+ * Use the `expandable slot` to add content that only shows up on click. This slot should not contain any relevant action since it is only shown on click.
  *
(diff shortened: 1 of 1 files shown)


packages/docs/src/stories/components/teaser-media.stories.ts
```diff
@@ -178,7 +178,7 @@ export const MetaSlot = {
 };
 
 /**
- * Use the `expandable` slot to add content that only shows up on hover.
+ * Use the `expandable slot` to add content that only shows up on hover. This slot cannot contain any action since it is not shown on mobile devices due to hover interaction.
```
mariohamann: Again: The goal was to explicitly make it possible to "click", either with "Enter" or with a click – no need for only hover. Please re-clarify with @MartaPintoTeixeira  and think properly about the initial goal: https://github.com/solid-design-system/solid/pull/2645#discussion_r2695712940

MartaPintoTeixeira: in Expandable docs section
Code-only: Use the “expandable slot” to add content that only shows up on click. This slot should not contain any relevant action since it is only shown on click.

and

in Accessibility Information change this one to this:
Avoid placing essential information or actions in expandable teaser areas since they are only visible on click.

balco0110: i've changed the description

PR #2759, opened 2026-01-22
fix: disabled removabled when selected or toggleable

martascorreia:
## Description:
Closes: https://github.com/solid-design-system/solid/issues/2398.
Previous PR: https://github.com/solid-design-system/solid/pull/2682

Added a variable _isRemovable_ to the sd-tag component. This variable represents the _removable_ attribute being enabled, while checking if the  _toggleable_ or _selected_ attributes are disabled. Then, replaced every place that checked if the removable attribute was enabled for this new variable.

Also added a note to the docs about this. Let me know if it is enough or if we should create more emphasis. 

## Definition of Reviewable:

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,5 +1,21 @@
 # @solid-design-system/components
 
+## 6.3.0
+
+### Minor Changes
+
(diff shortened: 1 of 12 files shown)


packages/docs/src/stories/components/tag.test.stories.ts
```diff
@@ -115,6 +115,28 @@ export const Disabled = {
   }
 };
 
+export const RemovableAndSelectedAndToggleable = {
+  name: 'RemovablexSelectedxToggleable',
```
smfonseca:
```suggestion
  name: 'Removable x Selected x Toggleable',
```

PR #2760, opened 2026-01-22
fix: implementing close trigger + fixing tooltip not closing outside of element 

martascorreia:
## Description:
Closes https://github.com/solid-design-system/solid/issues/2295. 
Previous PR: https://github.com/solid-design-system/solid/pull/2661

A new attribute "close-trigger" was added to the tooltip, mirroring the current attribute "trigger" that now works more as a "open-trigger" (I didn't change the name, should I?). This "close-trigger" defines which triggers should make the tooltip close. For the specific case of sd-range where the tooltip should always stay open, the sd-range changes the "close-trigger" to manual.

## Definition of Reviewable:

### DIFF
--- a/.changeset/lucky-parks-raise.md
+++ b/.changeset/lucky-parks-raise.md
@@ -2,4 +2,4 @@
 '@solid-design-system/components': minor
 ---
 
-Implementing a 'close-trigger' attribute to better control what triggers closing the 'sd-tooltip'. Fixed tooltip not closing outside the element.
+Implementing a `close-trigger` attribute to better control what triggers closing the `sd-tooltip`. Fixed tooltip not closing outside the element.


.changeset/lucky-parks-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/components': minor
+---
+
+Implementing a 'close-trigger' attribute to better control what triggers closing the 'sd-tooltip'. Fixed tooltip not closing outside the element.
```
smfonseca:
```suggestion
Implementing a `close-trigger` attribute to better control what triggers closing the `sd-tooltip`. Fixed tooltip not closing outside the element.
```

Issue #2796, opened 2026-02-19
fix: 🤔 sd-step inside sd-step-group: disabled attribute is overridden by waiting


jwahler:
## Current behavior

When using `sd-step` inside an `sd-step-group`, manually setting the disabled attribute on a step does not persist. If a step has the disabled attribute explicitly defined, the `sd-step-group` logic removes disabled and instead applies the waiting attribute automatically.
As a result, manually configured state is overridden by the internal state management of the `sd-step-group`.

## Expected behavior

Manually defined attributes (e.g., disabled) must not be overridden by the `sd-step-group`.
It should be possible to individually control the enabled/disabled state of each `sd-step`, regardless of the group’s internal state logic.
**Automatic state management should respect explicitly defined attributes and only apply default states when no manual configuration is present.**

## Steps to reproduce

[Example on codepen.io](https://codepen.io/JoeWa/pen/QwEoYGJ)

1. Create an `sd-step-group`.
2. Add multiple `sd-step` components inside it.
3. Manually add the disabled attribute to one step.
4. Render the component.

Observe that:

1. The disabled attribute is removed.
2. The waiting attribute is applied instead.

## Technical Information

* Component: `sd-step`
* Parent Component: `sd-step-group`
* Version: 6.7.0
* Issue Type: Attribute override / state management conflict
* Impact: Prevents granular and programmatic control of individual step states

The component appears to automatically synchronize step states within the group without checking whether a state was explicitly defined by the developer.

## DoR
- [ ] Item has business value
- [ ] Item has been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

PR #2859, opened 2026-03-13
feat: adding manual management of sd-step-group states

martascorreia:
## Description:
Closes: https://github.com/solid-design-system/solid/issues/2796

A new manual management functionality was added to the `sd-step-group`, activated by setting the `active-step` attribute to -1. Rather then adding more conditions to cater to use cases like https://github.com/solid-design-system/solid/issues/2796, this functionality allows devs to control all states of `sd-step` inside the `sd-step-group` while still using the `sd-step-group` structure.

## Definition of Reviewable:


packages/components/src/components/step-group/step-group.ts
```diff
@@ -103,8 +112,10 @@ export default class SdStepGroup extends SolidElement {
           step.current = true;
         } else if (i > index) {
           step.waiting = true;
+          step.current = false;
         } else {
           step.current = false;
+          step.waiting = false;
```
mariohamann, line 118: What did this change do?

PR #2889, opened 2026-04-07
feat: Added realistic template for sd-tag

### LINKED ISSUE #2134: feat[dev]: ✨ Realistic sd-tag template
## Description
We have updated the sd-tag template with real content and we need code to update it in storybook to keep consistency.

FIgma: https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Component-Docs?node-id=3274-20467&t=vs5GMMxS8eVFpyN4-4

- [ ] Question1
- [ ] Question2

## DoR
- [ ] Item dependencies have been identified and documented

## DoD

balco0110:
## Description:

Closes: https://github.com/solid-design-system/solid/issues/2134
## Definition of Reviewable:


MartaPintoTeixeira (commented):
This things are not as in figma:
- Remove "Risk" in each tag of the section "Filter Tag Group with Risk"
- "Moderate" instead of "Moderated" in section "Filter Tag Group with Risk"
- In this section "Removable Filter Tag Group" the underline of "Remove..." is starting before the R (it has one space more then it should)
- In this section "Removable Filter Tag Group" in figma the space between "ETFs" tag and "Remove filters" is bigger then in code. in figma "Remove filters has a left padding of 16. 
- in this section "Tag Group with Links" tags are suposed to hug horizontally the content as in figma. they should not have all the same horizontal size.

Assign me again once solved. I will remove my assignment

MartaPintoTeixeira (dismissed):

martascorreia (approved):

martascorreia (dismissed):
The sd-tag overview page references the template "Tag". I think for consistency, it should reference the different stories "Filter Tag Group" "Filter Tag Group with Morningstart Rating" etc, right?

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
+        </sd-tag>
+        <sd-tag toggleable>
+          <label class="sr-only">1 star</label>
+          <sd-icon name="system/star-filled" color="currentColor"></sd-icon>
+        </sd-tag>
+      </div>
+    </div>
+
+    <script type="module">
+      const group = document.querySelector('.morningstar-tag-group');
```
mariohamann: Use id instead of class

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
-    </sd-tag>
-    <sd-tag toggleable>
-      <label class="sr-only">1 star</label>
-      <sd-icon name="system/star-filled" color="currentColor"></sd-icon>
-    </sd-tag>
+    <div>
+      <p class="sd-paragraph sd-paragraph--size-lg font-bold mb-4">
+        Find Top-Rated Investments with Morningstar Ratings:
+      </p>
+      <div class="morningstar-tag-group flex gap-2">
```
mariohamann: Semantically I'm not sure if this is correct to be honest. Apparently these are exclusive buttons, which is not transported in any way, that others are decactivated. Can you please inform yourself, what the best approach would be here to make this correct from a11y perspective?

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
+          Low
+        </sd-tag>
+      </div>
+    </div>
+
+    <script type="module">
+      const group = document.querySelector('.risk-tag-group');
+      const tags = group.querySelectorAll('sd-tag');
+
+      tags.forEach(tag => {
```
mariohamann: Can't we jsut set even listener on the group and then check inside the event if the click was a tag? Just to reduce the amount of event listeners.

auroraVasconcelos:
> The sd-tag overview page references the template "Tag". I think for consistency, it should reference the different stories "Filter Tag Group" "Filter Tag Group with Morningstart Rating" etc, right?

Hey @martascorreia, when linking a template, use the component’s name if it’s the main component of that template (like in this case). Only use the story name when the template includes multiple components and the one you’re linking is not the main one.

You can find more details about documenting in Storybook here: https://solid-design-system.fe.union-investment.de/docs/?path=/docs/packages-docs-contributing--docs

packages/docs/src/stories/templates/tag.stories.ts
balco0110: Done, updated all three tag group stories to use id instad of class

packages/docs/src/stories/templates/tag.stories.ts
balco0110: I've added role="group" with an aria-label to the container for all of the three toggle stories, which communicates the mutual exclusivity to screen readers. Maybe we should consider to do some component level change to sd-tag to add role="radio".

packages/docs/src/stories/templates/tag.stories.ts
balco0110: i've refactored the stories to use a single event listener on the group container, checking the closest tag to confirm the click target

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
-      <sd-tag removable>Filter 4</sd-tag>
+    <div>
+      <p class="sd-paragraph sd-paragraph--size-lg font-bold mb-4">Active filters:</p>
+
+      <div id="removable-tag-filters" class="flex items-center gap-2 flex-nowrap">
+        <sd-tag removable>Stocks</sd-tag>
+        <sd-tag removable>Bonds</sd-tag>
+        <sd-tag removable>Mutual funds</sd-tag>
+        <sd-tag removable>ETFs</sd-tag>
+        <sd-link
```
smfonseca: We should use `sd-button` in this case since `remove filters` is an action, not navigation.

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
-      <sd-tag toggleable>
-        <div class="h-4 w-4 bg-risk-low border-primary-800 border-[1px]"></div>
-        Low Risk
-      </sd-tag>`
+  render: () => html`
+    <div>
+      <p class="sd-paragraph sd-paragraph--size-lg font-bold mb-4">Risk level:</p>
+      <div id="risk-tag-group" class="flex gap-2">
+        <sd-tag toggleable selected>
+          <div class="h-4 w-4 border-primary-800 border-[1px] bg-[var(--fill-risk-veryhigh)]"></div>
```
smfonseca: Have you checked if it's possible to use the bg-class instead of setting it like this?

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
-    <sd-tag toggleable>
-      <label class="sr-only">1 star</label>
-      <sd-icon name="system/star-filled" color="currentColor"></sd-icon>
-    </sd-tag>
+    <div>
+      <p class="sd-paragraph sd-paragraph--size-lg font-bold mb-4">
+        Find Top-Rated Investments with Morningstar Ratings:
+      </p>
+      <div id="morningstar-tag-group" class="flex gap-2" role="group" aria-label="Filter by morningstar rating">
+        <sd-tag aria-checked="true" toggleable selected>
```
smfonseca: You need to update `aria-checked` when clicking on the tags and it should have `role="radio"`

packages/docs/src/stories/templates/tag.stories.ts
```diff
@@ -12,94 +12,169 @@ export default {
...
-    </sd-tag>
-    <sd-tag toggleable>
-      <label class="sr-only">1 star</label>
-      <sd-icon name="system/star-filled" color="currentColor"></sd-icon>
-    </sd-tag>
+    <div>
+      <p class="sd-paragraph sd-paragraph--size-lg font-bold mb-4">
+        Find Top-Rated Investments with Morningstar Ratings:
+      </p>
+      <div id="morningstar-tag-group" class="flex gap-2" role="group" aria-label="Filter by morningstar rating">
```
smfonseca: Here it should be `role="radiogroup"` because it communicates that only one option can be selected.

smfonseca (changes requested):

smfonseca (approved):

PR #2912, opened 2026-04-14
chore: update dependencies [skip chromatic]

rahulsuvarna18:
## Description:
closes:https://github.com/orgs/solid-design-system/projects/1/views/72?pane=issue&itemId=51149554&issue=solid-design-system%7Csolid%7C460

## Definition of Reviewable:


auroraVasconcelos (commented):
Please add a changelog

auroraVasconcelos (commented):
The root didn't had packages outdated?

auroraVasconcelos (dismissed):

rahulsuvarna18:
Thanks for the tip. As requested, I have added the missing root file dependencies and the changeset.

rahulsuvarna18:
Vite was not updated. Here is the follow up ticket for it
https://github.com/orgs/solid-design-system/projects/1/views/72?pane=issue&itemId=175694429&issue=solid-design-system%7Csolid%7C2910

.changeset/two-beds-go.md
```diff
@@ -0,0 +1,9 @@
+---
```
smfonseca, line 1: If we now follow the rule that we add a changeset for updated `dependencies`, `styles` and `tokens` package should be removed from this changeset since only `devDependencies` in these packages have been changed.

Also, I feel it would make sense to list the updated dependencies for users to easily identify them.

mariohamann: As an alternative we could just list all changed dependencies and add the link to the PR instead of listing decDependencies...?

smfonseca: Sure, that could also be an option.

smfonseca:
@rahulsuvarna18 please use this PR to update `vite` to version `7.3.2` so that we can close this [PR](https://github.com/solid-design-system/solid/pull/2886) from renovate bot.

After doing the update here, you can close the renovate bot PR and add a comment pointing to this PR.

smfonseca (approved):

PR #2960, opened 2026-04-28
fix: checkbox, switch and radio focus outline color not applied

### LINKED ISSUE #2733: fix: 🤔 Checkbox focus outline color not applied (only with multi-theming)
## Current behavior
When navigating through the checkbox using the keyboard, the focus ring (focus outline) appears dark instead of using the theme color.

Image

## Expected behavior
When navigating through the checkbox using the keyboard, the focus outline / focus ring should use the theme primary color #0066B3.

## Steps to reproduce
Go to the Checkbox or Checkbox Item documentation page (only reproducible with multi-theming).
Navigate through the component using the keyboard (Tab / Shift + Tab).
Observe the focus outline color.

## Technical information
The correct theme variables are being resolved, however the value is not wrapped in the required rgb() function, causing the browser to fall back to the default focus outline color.

Image

## DoR
- [ ] Item has business value
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/2733
## Definition of Reviewable:


packages/components/src/components/radio/radio.ts
```diff
@@ -190,6 +190,11 @@ export default class SdRadio extends SolidElement {
         @apply outline-none;
       }
 
+      [part='control--checked'],
```
smfonseca, line 193: This can't be applied directly in the parts?

tega-ekakitie:
@balco0110 looks good to me

packages/components/src/components/radio/radio.ts
balco0110: radio has no underlying <input> element — focus lives directly on the host. Using :host(:focus-visible) [part] in CSS ensures the outline only appears on keyboard navigation (not on mouse clicks), which is the correct behavior. Applying it via hasFocus would show the ring on mouse clicks too.

.changeset/mighty-goats-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/tokens': patch
+---
+
+Wrapped `--outline-color-primary`, `--outline-color-error` and `--outline-color-white` in `rgba()` so `outline-*` utilities resolve correctly in multi-theming
```
mariohamann: Describe what this solves the user. THis is a fix for TailwindCSS users in aprticular.

.changeset/tangy-islands-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@solid-design-system/components': patch
+---
+
+Fixed focus outline color in `sd-checkbox`, `sd-switch` and `sd-radio` not using the primary color.
```
mariohamann: Describe the fix not what didn't work before

.changeset/mighty-goats-sleep.md
balco0110: i've updated the changeset

.changeset/tangy-islands-scream.md
balco0110: the changeset is updated

martascorreia (approved):

PR #2965, opened 2026-04-29
[no ci] fix(2057): undefined tailwind variables in production, 

### LINKED ISSUE #2057: fix: undefined tailwind variables in production
## Current behavior
We recently received two bugs reports that are connected with tailwind variables not being defined.

- sd-copyright: the default shadow was not being displayed because tw-blur, tw-sepia, tw-brightness, tw-contrast and others were undefined and this broke the filter css effect. [Fix](https://github.com/solid-design-system/solid/pull/2025)
- sd-teaser-media: the gradients were not being displayed because the start and end position variables were undefined. [Fix](https://github.com/solid-design-system/solid/pull/1810)

In Storybook these problems were not identified because Tailwind is loaded and the classes are defined. The issue is visible when testing on Codepen or other platforms that do not have Tailwind included.

For now @mariohamann and I agreed the following:

- for styles components the missing variables should be added in the the respective component css.
- for js components the missing variables should be added in `solid-element.ts` under `:host`.

This approach keeps the variables scoped to the respective components.
If we see that many other components start having this kind of issue, then we need to look for another solution. 

The bigger issue is that our screenshot tests are not reliable enough because they are also not detecting this kind of problem since Tailwind classes are also present there.

## Expected behavior
Screenshot tests should replicate similar conditions to production and fail if something is not right with the components.

## Steps to reproduce
- remove `--tw` classes from `copyright.css`
- build SDS
- open component in codepen and replace urls to use local builds for components and styles
- shadow should be missing from the copyright

## Technical Information

A possible solution could be to remove Tailwind from our screenshot tests, maybe with a class `tailwind-not-allowed` or similar that would strip all tailwind classes from the tests as if it was a production environment. The still needs to be investigated more.

We have also seen some reports that cssnano removes empty variables during minification, so this could also be a point to investigate.

## DoR
- [x] Item has business value

## DoD

clemens-vi:
NOTE: implemented by dev-bufgix.agent

## Description:
This pull request addresses a production bug (#2057) where Tailwind CSS gradient and shadow variables were undefined, causing rendering issues for shadow overlays in table cells and hidden links. The changes ensure these variables are always defined, even when Tailwind is not loaded, and add a Storybook test to verify the fix.

**Bug fixes for shadow and gradient CSS variables:**

* [`packages/styles/src/modules/table-cell.css`](diffhunk://#diff-ddb538f5e1fba6f37b0e5373335359f9ec8f36b885e372f9a185a33d45e237aeR19-R25): Defines default values for all `--tw-gradient-*` CSS variables in `.sd-table-cell` to prevent them from being undefined in production environments without Tailwind.
* [`packages/styles/src/modules/hidden-links.css`](diffhunk://#diff-62859e38f1e680731ec1fb8df5f6544689324d24ec45bee44c127725b582d422R24-R29): Sets default values for all `--tw-shadow*` CSS variables in `.sd-hidden-links--multiple` for the same reason.

**Testing and documentation:**

* [`packages/docs/src/stories/styles/table-cell.test.stories.ts`](diffhunk://#diff-d047612276665a1b2fff7e298df2783f29cb0a2cf988caf1bdb469c00491c396L51-R103): Adds a new `Shadow` Storybook story that demonstrates and verifies the correct rendering of shadow overlays on table cells, ensuring the bug is fixed and variables are present.
* [`packages/docs/src/stories/styles/table-cell.test.stories.ts`](diffhunk://#diff-d047612276665a1b2fff7e298df2783f29cb0a2cf988caf1bdb469c00491c396R9): Imports `html` from `lit-html` to support the new Storybook story rendering.

## Definition of Reviewable:

### DIFF
--- a/packages/docs/src/stories/styles/table-cell.test.stories.ts
+++ b/packages/docs/src/stories/styles/table-cell.test.stories.ts
@@ -6,6 +6,7 @@ import {
   storybookTemplate,
   storybookUtilities
 } from '../../../scripts/storybook/helper';
+import { html } from 'lit-html';
 
 const { argTypes, parameters } = storybookDefaults('sd-table-cell');
 const { overrideArgs } = storybookHelpers('sd-table-cell');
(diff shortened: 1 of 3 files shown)


mariohamann:
Not happy with the solution. Would pick that up later again.

Issue #1059, opened 2024-05-08
feat: ✨ add testing and tracking attributes to components


karlbaumhauer:
## User Story
As a user of the Solid Design System, I would like to utilize an additional attribute to be used for tracking implemenatation, as well as an additional attribute to be used for testing on each component, so that solid components can be used with the current tracking implementation and within the current testing structure.
### Suggested Solution
A global script adds the following attributes to each component within the build process:
1. Tracking 
key: data-sd-version
value: current version of the library (eg. 3-3-0)
2. Testing
key: data-sd-name
value: component name (eg. `sd-button`)

## Subtasks
- [ ] Design_implementation_task
- [ ] Development_implementation_task
- [ ] Design_release_task

## DoR
- [ ] Item has business value
- [ ] Item all subtasks have been estimated by the team
- [ ] Item is clear and well-defined
- [ ] Item dependencies have been identified

## DoD
- [ ] All subtasks have been closed

karlbaumhauer:
@mariohamann assigned to you in order to read through it and see if anything is missing. I could not find any information for the tracking part and especially no information about the priority.

@yoezlem assigned to you in order to prioritize. I did not create it as epic, as we only have dev implementation for it. The testing prio is medium IMO.

mariohamann:
Updated and removed myself.

karlbaumhauer:
@yoezlem we need to clarify if this is still a valid request. The Jira Ticket is still open.

karlbaumhauer:
> [@yoezlem](https://github.com/yoezlem) we need to clarify if this is still a valid request. The Jira Ticket is still open.

@mariohamann / @yoezlem I added some context, I received from the testing team, to the description.

PR #3114, opened 2026-06-29
feat: expose ./package.json in exports

SchulteDev:
## Description:

Add `"./package.json": "./package.json"` to the `exports` map of the published packages so downstream consumers can resolve and read the package's own `package.json` — most usefully to read the installed `version` — programmatically.

- `packages/components/package.json` — added, placed right after the existing `"."` entry.
- `packages/tokens/package.json` — added as the first `exports` entry (this package has no `"."` entry).
- `packages/styles/package.json` — already had this entry; unchanged.

No other exports were touched; there is no version-constant export and no build changes. A patch changeset is included.

### Why

Today the published `exports` map gates subpath access and there is no `./package.json` export, so a consumer hits `ERR_PACKAGE_PATH_NOT_EXPORTED`:

```js
require.resolve('@solid-design-system/components/package.json') // throws today
```

This complements #1059 — which surfaces the version only as a rendered `data-sd-version` DOM attribute via build-time injection — by giving **programmatic** access to the version rather than a DOM-only attribute. The two approaches are independent and address different use cases (runtime/tooling reads vs. inspecting rendered markup).

Re-exposing `package.json` through `exports` is a well-established convention (React and many other libraries do exactly this). It exposes no new information — the published manifest is already public on the npm registry; this only makes it resolvable through Node's module-resolution algorithm.

### Verification

Each touched package was packed (`npm pack`) and installed into a clean local project from the tarballs:

- `require.resolve('@solid-design-system/components/package.json')` succeeds (CJS) — same for `tokens` and `styles`.
- `import.meta.resolve('@solid-design-system/components/package.json')` succeeds (ESM) — same for `tokens` and `styles`.
- Importing the JSON and reading `.version` returns the current version.
- Negative control: an unlisted subpath (e.g. `.../README.md`) still correctly throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, confirming the gate is unchanged apart from the added entry.
- `npm pack --dry-run` file lists are identical before and after for both touched packages — packed output is otherwise unchanged.

## Definition of Reviewable:
- [x] No documentation changes required (purely additive `exports` entry; behaviour for existing subpaths is unchanged)
- [x] No migration guide required (non-breaking, additive)
- [x] No E2E/visual changes (no component, story, or rendered behaviour is affected)
- ~~Stories created/updated~~ — N/A
- Related: #1059

### DIFF
--- a/.changeset/expose-package-json-in-exports.md
+++ b/.changeset/expose-package-json-in-exports.md
@@ -3,4 +3,4 @@
 "@solid-design-system/tokens": minor
 ---
 
-Expose `./package.json` in the package exports so you can read the installed version programmatically, for example `import pkg from '@solid-design-system/components/package.json' with { type: 'json' }` or `require.resolve('@solid-design-system/components/package.json')`. Previously the exports map blocked this subpath.
+Expose `./package.json` in the package exports so you can read the installed version programmatically, for example `import pkg from '@solid-design-system/components/package.json' with { type: 'json' }` or `require.resolve('@solid-design-system/components/package.json')`.
--- a/.changeset/expose-package-json-in-exports.md
+++ b/.changeset/expose-package-json-in-exports.md
@@ -1,6 +1,6 @@
 ---
-"@solid-design-system/components": patch
-"@solid-design-system/tokens": patch
+"@solid-design-system/components": minor
+"@solid-design-system/tokens": minor
 ---
 
 Expose `./package.json` in the package exports so you can read the installed version programmatically, for example `import pkg from '@solid-design-system/components/package.json' with { type: 'json' }` or `require.resolve('@solid-design-system/components/package.json')`. Previously the exports map blocked this subpath.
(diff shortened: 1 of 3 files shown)


.changeset/expose-package-json-in-exports.md
```diff
@@ -0,0 +1,6 @@
+---
+"@solid-design-system/components": patch
+"@solid-design-system/tokens": patch
```
mariohamann:
```suggestion
"@solid-design-system/components": minor
"@solid-design-system/tokens": minor
```

mariohamann:
Thanks for your contribution! Please adapt the changeset, as this a feature and is not fixing a bug – besides that I'm fine with that!

.changeset/expose-package-json-in-exports.md
SchulteDev: Thanks for the review! Applied — the changeset now declares a `minor` bump for both packages, since exposing a new `./package.json` export is an additive feature rather than a fix. I also trimmed a redundant "previously blocked" history note so the changeset focuses on the consumer-facing value.

karlbaumhauer:
Due to limitations of tokens to our repo, this fork-PR cannot be merged. The changes will be handeled in #3119

PR #3131, opened 2026-07-07
docs: update Storybook Introduction page

### LINKED ISSUE #3044: feat[dev]: Update Introduction Page on Storybook
## User Story
As a user of Solid Ds I would expect the [General Introduction](https://solid-design-system.fe.union-investment.de/docs/?path=/docs/docs-general-introduction--docs) page in storybook to:

- have a short intro about Union Investment
- show Union Investment logo
This way the user will quickly connect Solid with Union Investment.

Additionally (dev only):
There is a link in the wrong color. it should be primary blue:
"([tested Browser versions](https://github.com/microsoft/playwright/releases/tag/v1.60.0))"

### Suggested Solution

H1: "Introduction"

Row:
UI Logo (left)
Text (right) "With over 6 million clients, Union Investment is one of Germany’s leading fund management companies. As part of the Genossenschaftliche FinanzGruppe, we offer investment solutions for private clients, institutional clients, and real estate clients in a constantly changing environment."

H2: "Solid Design System"

Then the vision part below ...

--

Change vision text:

"We are the sustainable foundation for all digital products that strengthens the brand, is user-friendly and accessible. We enable and guide designers, developers and agents to use it productively and with confidence."

## DoR
- [x] Item has business value

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/3044
## Definition of Reviewable:


MartaPintoTeixeira (dismissed):

packages/docs/src/stories/docs/general/01-Introduction.mdx
```diff
@@ -49,7 +55,7 @@ Solid Components are designed to offer high reliability and easy maintainability
 - **"Dumber", declarative components**: Components free of business logic for easier updates.
 - **Cautious feature implementation**: New features are carefully evaluated for their impact on reliability and maintainability.
 - **Responsibility delegation to the application**: Lightweight components allowing for customization and greater flexibility.
-- **Thorough testing**: High test coverage with E2E-Tests via Playwright (<a href={`https://github.com/microsoft/playwright/releases/tag/v${playwrightVersion}`} target="\_blank">tested Browser versions</a>) and Visual Regression Tests via Chromatic ([current Browser versions](https://www.chromatic.com/docs/infrastructure-release-notes)).
+- **Thorough testing**: High test coverage with E2E-Tests via Playwright (<a href={`https://github.com/microsoft/playwright/releases/tag/v${playwrightVersion}`} target="\_blank" className="underline! transition-colors text-primary! hover:text-primary-500! active:text-primary-800!">tested Browser versions</a>) and Visual Regression Tests via Chromatic ([current Browser versions](https://www.chromatic.com/docs/infrastructure-release-notes)).
```
auroraVasconcelos: This line uses two different approaches for links: one uses HTML with inline styles, while the other uses Markdown. Could we use the same approach for both to keep things consistent?

IF we keep the HTML version, are the inline styles actually needed? Would using our sd-link component or the sd-interactive classes make sense here?

balco0110: Switched both to <sd-link>, same approach for both now, and no inline styles needed since the component handles the styling itself.

auroraVasconcelos (approved):

MartaPintoTeixeira (approved):

Issue #3075, opened 2026-06-11
fix: clean up unused CSS variables (BREAKING CHANGES)


auroraVasconcelos:
During the multi-theming QA process, several variables were removed from both the codebase and Figma to improve consistency and maintainability. This introduces a breaking change, as the latest version of the tokens package is required. Furthermore, if themes are being overridden, some CSS variables may need to be updated or replaced by the users.

The following CSS variables should be removed from the code.

## Technical Information
This change was communicated to users and documented in the following resources. Useful documentation:
- https://solid-design-system.fe.union-investment.de/docs/?path=/docs/packages-tokens-installation--docs#%EF%B8%8F-upcoming-breaking-changes
- https://teams.microsoft.com/l/message/19:sycqZKo6K_-P5eYh0H4Xn2UpoKmti8Ar5SICMb9z7C41@thread.tacv2/1775829718512?tenantId=2eb07b91-f9f8-437b-a89e-59f55e565c11&groupId=cec93f98-ceee-42bd-bfde-a6b51fc1194e&parentMessageId=1775829718512&teamName=Solid%20Design%20System%20Community%20(Team)&channelName=General&createdTime=1775829718512

- CSS code that needs to be cleaned up is marked with comments starting with TODO, along with an explanation of what needs to be changed.
- The variables are documented as Deprecated on the component documentation pages with the message: "This custom property is deprecated." The replacement variables are also documented.

Additional theme with legacy variables for e.g. CMS team needed or properly document which ones have to be inserted?

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

martascorreia:
main change is in this commit: https://github.com/solid-design-system/solid/pull/3240/changes/9387dc4da20774623afb0af7ffcc34fc03f4fc9e

PR #3179, opened 2026-07-22
feat: tokens stability for breaking changes

### LINKED ISSUE #2764: feat: ensure tokens stability for breaking changes
## User Story

Apps should not integrate CSS variables into their own app, they should always be provided from outside, otherwise theming is getting impossible.

Problem: When we in the future remove USED variables, this could break apps, when the surrounding environment doesn't provide the variable anymore.

Introduce adding the UI variables as a fallback even if the themes css is not loaded.

### Suggested Solution
We always provide compatibility CSS, that can be added by the surrounding app, if it's aware of apps using the variable – which actually should always be taken, when the apps are not under own control or can be properly tested.

## DoR
- [x] Item has business value

## DoD

martascorreia:
## Description:
closes https://github.com/solid-design-system/solid/issues/2764

## Definition of Reviewable:


martascorreia:
relates to https://github.com/solid-design-system/solid/issues/3075 as this branch already removes the automatic adding of legacy variables to the themes' stylesheets.

.changeset/rich-teeth-prove.md
```diff
@@ -0,0 +1,7 @@
+---
+'@solid-design-system/tokens': minor
+'@solid-design-system/docs': minor
```
mariohamann: Major? :)

packages/tokens/themes/sp/sp.css
```diff
@@ -523,33 +522,4 @@
...
-  --sd-menu--disabled-color-icon: var(--sd-menu-item--disabled-color-icon-fill);
-  --sd-menu--disabled-color-text: var(--sd-menu-item--disabled-color-text);
-  --sd-tag-font-weight: var(--sd-choice-control-font-weight);
-  --sd-tag--disabled-color-border: var(--sd-color-border-neutral-500);
-  --sd-navigation-item--current-color-text: var(--sd-navigation-item-color-text);
-  --sd-option--disabled-color-border: var(--sd-color-border-neutral-500);
-  --sd-audio__timestamp-color-text: var(--sd-color-text-neutral-700);
-  --sd-accordion__chevron-color-text: var(--sd-accordion-color-icon-fill, var(--sd-color-icon-fill-primary));
-  --sd-form-control__tooltip-color-icon: var(--sd-interactive--default-color-text);
 }
```
mariohamann, line 555: As these are removed it's now a Breaking Change.

packages/docs/src/stories/packages/tokens/Usage.mdx
```diff
@@ -122,88 +122,114 @@ This will only affect the primary background color inside `sd-button` components
 
 ### Deprecated CSS variables (tokens)
 
-The following CSS variables (tokens) are deprecated and will be removed in a future release. Update your custom styles to use the recommended replacements.
+The following CSS variables (tokens) are deprecated and will be removed in a future release. Please update your custom styles to use the recommended replacements.
```
mariohamann: They ARE removed now, right? :)

packages/tokens/themes/tailwind.css
```diff
@@ -8,178 +8,178 @@
...
-  --text-xs: var(--sd-text-xs);
+  --spacing: var(--sd-spacing-1, 0.25rem);
+  --text-2xl: var(--sd-text-2xl, 1.75rem);
+  --text-3xl: var(--sd-text-3xl, 2rem);
+  --text-4xl: var(--sd-text-4xl, 2.5rem);
+  --text-base: var(--sd-text-base, 1rem);
+  --text-color-accent: rgba(var(--sd-color-text-accent, 45 157 0));
+  --text-color-black: rgba(var(--sd-color-text-black, 24 24 24));
+  --text-color-error: rgba(var(--sd-color-text-error, 204 25 55));
+  --text-color-neutral-500: rgba(var(--sd-color-text-neutral-500, 174 174 174));
```
mariohamann, line 168: As we now have fallbacks, we should definitely highlight to the users! Themes don't have to be installed anymore if there are only components using >v6 and higher or <v4 and lower.

packages/docs/src/stories/packages/tokens/Usage.mdx
martascorreia: They are removed from the theming sheets, but not yet from the components, that is https://github.com/solid-design-system/solid/issues/3075. So I only added the paragraph after, but I can already rephrase it as removed.

About this, do you think we should leave the list from now on and keep adding to it every time we have a new deprecated variable? I add the extra "removed" column to differentiate between the already removed (for this breaking change) and the ones that we could remove in the future, but I'm not sure if this will confuse users:
image

mariohamann: What if we just put it into the Breaking Changes?

PR #3236, opened 2026-08-06
feat: tokens stability

### LINKED ISSUE #2764: feat: ensure tokens stability for breaking changes

martascorreia:
## Description:
closes https://github.com/solid-design-system/solid/issues/2764

## Definition of Reviewable:


martascorreia:
Previous pr: https://github.com/solid-design-system/solid/pull/3179
Had to close it because of the limit of built branches issue we've been having. Once it fails, github-pages bugs and the easiest fix is opening a new PR.

Issue #3109, opened 2026-06-23
feat: Decouple drop shadow from the fixed attribute


coraliefeil:
## User Story

As a **consumer** of the Solid Design System, I would like the drop shadow on the fixed header to be optional, so that products can use a fixed header without enforced visual elevation, and can align the header appearance with their individual design language.

### Suggested Solution

Decouple the drop shadow from the `fixed` attribute. Instead of automatically applying the shadow when `fixed: true` is set, introduce a separate `shadow` attribute (or a `fixed-shadow`/`with-shadow` variant) that controls shadow rendering independently.

This allows consumers to combine:
- `fixed: true` + `shadow: true` → current elevated fixed header (default, backward-compatible)
- `fixed: true` + `shadow: false` → fixed header without drop shadow

### Technical Information

- Current behavior: `fixed: true` unconditionally applies a drop shadow to elevate the header above scrolled content
- Desired behavior: shadow is opt-in (or opt-out), decoupled from the `fixed` attribute
- Check if shadow is currently applied via a CSS class, a token, or inline style – adjust accordingly
- Ensure the default remains backward-compatible (shadow still appears by default when `fixed: true`)

## DoR
- [x] Item has business value
- [x] Item has been estimated by the team
- [x] Item is clear and well-defined
- [x] Item dependencies have been identified

## DoD
- [x] Brand approval (if neccessary)
- [x] Figma documentation file incl. hand-off has been created/updated (if applicable)
- [x] Tokens have been applied/updated (if applicable)
- [x] Crosslinks are created/updated
- [x] Implementation works successfully on `figma` branch

coraliefeil:
@solid-design-system/development who can review [figma branch](https://www.figma.com/design/LJcrIgF1jFGDT6SbY6fwSb/-3109-decouple-header-drop-shadow?node-id=6646-28795) ?

auroraVasconcelos:
Everything seems fine

coraliefeil:
I’m putting this one back to ready ... for dev implementation within the same ticket. Who can do that within the current breaking changes sprint? @solid-design-system/development

PR #3246, opened 2026-08-11
feat: decouple drop shadow from the fixed attribute

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/3109
## Definition of Reviewable:


packages/docs/src/stories/packages/components/Migration/from-v6.mdx
```diff
@@ -134,3 +134,40 @@ Both components previously implemented their own modal behavior (a custom `Modal
 ```html
 <sd-dialog open>...</sd-dialog> <sd-drawer open>...</sd-drawer>
 ```
+
+## `sd-header` drop shadow is no longer tied to the `fixed` attribute
+
+Previously, setting `fixed` on `sd-header` always rendered a drop shadow underneath it. The shadow is now controlled by a separate `shadow` attribute, which is off by default and can be combined with `fixed` or used on its own.
+
+**What changed and why:** Position and elevation were coupled in a single attribute, so products that needed a fixed header had to accept the enforced elevation. Separating them lets consumers align the header with their own design language.
```
auroraVasconcelos: Do we really need this?

packages/docs/src/stories/packages/components/Migration/from-v6.mdx
```diff
@@ -134,3 +134,40 @@ Both components previously implemented their own modal behavior (a custom `Modal
...
+
+**Who is affected:** Anyone using `<sd-header fixed>`. The header keeps its fixed position, but renders no drop shadow until `shadow` is added.
```
auroraVasconcelos: Do we need to add this? Isn't it clear in the first pagraph?

packages/docs/src/stories/packages/components/Migration/from-v6.mdx
```diff
@@ -134,3 +134,40 @@ Both components previously implemented their own modal behavior (a custom `Modal
...
+
+**Migration steps:**
+
+- Add the `shadow` attribute to every fixed header that should keep its drop shadow.
+- If you previously suppressed the shadow by overriding `--sd-header__shadow-height` with `0`, that override is no longer needed and can be removed.
```
auroraVasconcelos: Could we use a more general approach?
For example: "If you previously suppressed the header shadow, the custom override is no longer needed and can be removed."

auroraVasconcelos (commented):
Although this issue specifically refers to the shadow, some themes (e.g. VB) use an underline instead. I assume it should behave in the same way as the UI theme’s shadow, but this behaviour is still missing. (Check the lines 124–128).

Image

balco0110:
> Although this issue specifically refers to the shadow, some themes (e.g. VB) use an underline instead. I assume it should behave in the same way as the UI theme’s shadow, but this behaviour is still missing. (Check the lines 124–128).
> 
> Image

i've updated the code, added this as a part from the shadow

van-nguyen-ht:
hi @balco0110 b I'm done with the review :), I left the changes for sd-teaser untouched since I'm not sure what the context there is.

packages/components/src/components/header/header.ts
```diff
@@ -23,13 +23,16 @@ import type { PropertyValues } from 'lit';
...
- * @cssproperty --sd-header__shadow-height - height of the shadow when the header is fixed
+ * @cssproperty --sd-header__shadow-height - height of the shadow
  */
 @customElement('sd-header')
 export default class SdHeader extends SolidElement {
-  /**  Determines whether the header is fixed or not. If the header is fixed at the top of the page, a shadow is shown underneath. */
+  /**  Determines whether the header is fixed or not. If the header is fixed, it stays at the top of the page while scrolling. */
   @property({ reflect: true, type: Boolean }) fixed = false;
 
+  /**  Determines whether a drop shadow is shown underneath the header. Can be used with out without the `fixed` attribute */
```
auroraVasconcelos:
```suggestion
  /**  Determines whether a drop shadow is shown underneath the header. */
```

I understand that this sentence makes sense in the context of this PR, but we don’t usually mention that attributes can be used together, as that is the expected behavior.

auroraVasconcelos (dismissed):
Please see my comment regarding the documentation. Other than that, LGTM.

auroraVasconcelos (approved):

PR #3327, opened 2026-09-10
feat: added new variant to `sd-interactive`

### LINKED ISSUE #3154: feat[dev]: update sd-interactive and refactor the closing button for sd-drawer, sd-dialog and sd-notification
## Description
This tasks describes the Code implementation for: #3121 

## DoR

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/3154
## Definition of Reviewable:


packages/components/src/components/dialog/dialog.ts
```diff
@@ -20,14 +19,12 @@ import SolidElement from '../../internal/solid-element';
...
  * @since 1.40.0
  *
- * @dependency sd-button
  * @dependency sd-icon
  * @dependency sd-theme-listener
  *
  * @slot - The dialog's main content.
  * @slot headline - The dialog's headline. Alternatively, you can use the `headline` attribute.
  * @slot footer - The dialog's footer, usually one or more buttons representing various options.
- * @slot close-button - The dialog's close button. Works best with `<sd-button>` and `<sd-icon>`.
```
auroraVasconcelos, line 30: I believe we should keep the slot, however, after checking the code I'm not sure if this was working. I would align with the team to check what is going on.

packages/docs/src/stories/styles/interactive.stories.ts
```diff
@@ -44,6 +44,24 @@ export const Default = {
   }
 };
 
+/**
+ * Use the `sd-interactive--*` class for alternative appearances:
```
auroraVasconcelos:
```suggestion
 * Use the `sd-interactive--*` classes for alternative appearances:
```

packages/docs/src/stories/styles/interactive.stories.ts
```diff
@@ -44,6 +44,24 @@ export const Default = {
...
+ *
+ * - Primary is the default variant
```
auroraVasconcelos: Lower case

packages/styles/src/modules/interactive.css
auroraVasconcelos, line 53: The secondary variable may have a different behaviour in inverted, please align/check with design

packages/components/src/components/dialog/dialog.ts
balco0110: Hey Aurora, thanks for checking this, I had a look and I don't think the slot was ever really there.

I couldn't find <slot name="close-button"> anywhere in the dialog, on main there's only headline, the default one and footer. So if you tried to use it nothing would show up. I also searched the repo and we are not using it. Seems like the doc line was added when the component was first created, but the slot itself never was.

So I think removing it is safe, since nothing actually changes. The close-button part is still there for styling, I just updated the text because it's a normal <button> now instead of an sd-button

packages/docs/src/stories/styles/interactive.stories.ts
balco0110: Done

packages/styles/src/modules/interactive.css
balco0110: i've added inverted style for the variant

packages/components/src/components/dialog/dialog.ts
balco0110: we will remove the documentation for the slot

MartaPintoTeixeira (changes requested):
Interactive is good.

For the other 3 components I would use the the primary variant (the default) so in mouse hover and clicks i expect color changes.
Please check that.

Assign me again once solved. Thanks

balco0110:
> Interactive is good.
> 
> For the other 3 components I would use the the primary variant (the default) so in mouse hover and clicks i expect color changes. Please check that.
> 
> Assign me again once solved. Thanks

this is now implemented like this

packages/styles/src/modules/interactive.css
```diff
@@ -7,9 +7,12 @@
  * @boolean sd-interactive--disabled Makes an element look disabled.
  * @boolean sd-interactive--inverted Inverts the colors of an element.
  * @boolean sd-interactive--reset Resets the default browser styles of e.g., a button.
- *
+ * 
+ * @variant { secondary } sd-interactive--variant-... The interactive color variant.
+ * 
  * @cssproperty --sd-interactive--active-color-text - The text color for interactive elements in active state.
  * @cssproperty --sd-interactive--default-color-text - The default text color for interactive elements.
+ * @cssproperty --sd-interactive--secondary-color-text - The text color for interactive elements in disabled state.
```
auroraVasconcelos: This description is the same as the next one, which makes it confusing.

auroraVasconcelos (dismissed):
LGTM! I just left one comment on the `sd-interactive` documentation. Please check it, and don’t forget to update the branch.

packages/styles/src/modules/interactive.css
balco0110: i've updated this

MartaPintoTeixeira (approved):

PR #3335, opened 2026-09-14
fix: update the sd-notification documentation

### LINKED ISSUE #3187: docs: fix sd-notification documentation
## Description
The sd-notification toast notification documentation currently includes internal properties that are not part of the component and only intended to storybook documentation to work properly.

Displaying these properties may confuse users and give the impression that they are supported or safe to use. The documentation should only expose properties that users can configure.

## Where to put the documentation
https://solid-design-system.fe.union-investment.de/docs/?path=/docs/components-sd-notification--docs#toast-notification
Sync with figma as well: https://www.figma.com/design/YDktJcseQIIQbsuCpoKS4V/Solid-DS-%E2%80%93-Component-Docs?node-id=2265-6173&p=f&t=rS4d80GN071iSn2o-0

## DoR
- [x] Item has business value

balco0110:
## Description:

Closes https://github.com/solid-design-system/solid/issues/3187
## Definition of Reviewable:


MartaPintoTeixeira (changes requested):
Last story has 2 titles. Can you check?
Assign me again once changed. Thanks

packages/docs/src/stories/components/notification.stories.ts
```diff
@@ -241,26 +245,28 @@ export const ToastNotification = {
...
- * Use the `data-notification-position` attribute with the value `bottom-center` to align the toast to the bottom center.
+ * Use the `toastStack` attribute with the value `bottom-center` to align the toast to the bottom center.
  */
 export const ToastBottomCenter = {
-  name: 'Toast Bottom Center',
+  name: 'Toast placement bottom center',
   tags: ['skip-playwright'],
   render: (_args: Record<string, any>) => {
     return html`
+      <!-- preview-ignore:start -->
```
martascorreia: where these intended as well? the preview-ignore:end is visible in both

balco0110: i removed them

martascorreia: What I meant is that the code it self "preview-ignore:end" was visible when I clicked "show code" on Storybook, so the hiding was not being effective for these two cases specifically where `<!--  preview-ignore:end -->` was used. 

I think the references to data-notification-type="info" and data-notification-position="top-right" in the sd-button still do need to be hidden or maybe the script slightly changed so that we don't have to use them?

martascorreia (approved):

auroraVasconcelos (approved):

MartaPintoTeixeira (approved):

